PR #1, opened 2020-08-13
Prototype Polina

polinasand:
Create servlet for keeping data about dishes and process it. Create fake database. Add search bar to find dishes.


src/main/java/com/google/sps/servlets/SearchMealServlet.java
```diff
@@ -0,0 +1,64 @@
+package com.google.sps.servlets;
```
grenlayk: Please add Apache header to all files containing source code, as it mentioned [here](https://github.com/google/new-project).

src/main/webapp/script.js
```diff
@@ -0,0 +1,17 @@
+function searchDishes() {
+    fetch('/search-dishes').then(response => response.json()).then((dishes) => 
+    {
+        let container = document.getElementById("dishes-container");
+        container.innerText = "";
+        dishes = dishes || [""];
```
grenlayk: Nit:
I suppose, it's better to use `??` operator instead of `||` in this case.
[MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) with example and description.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
```diff
@@ -0,0 +1,64 @@
...
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        result = searchDishByKeyWord(getParameter(request, "text", ""));
+        
+        response.setContentType("application/json;");
+        response.getWriter().println(result);
+        response.sendRedirect("/index.html");
+    }
+    
+    private ArrayList<String> searchDishByKeyWord(String request) {
+        ArrayList<String> result = new ArrayList<String>();
```
grenlayk: Nit:
I think you can use [`.filter()`](https://www.baeldung.com/java-stream-filter-lambda) method here.

grenlayk (approved):

polinasand:
@googlebot I fixed it.

zajonc-gugle:
@googlebot I fixed it.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
bprusak: Please hit "resolve" on any comment you already addressed, such as this one.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.gson.Gson;
+
+@WebServlet("/search-meal")
+public class SearchMealServlet extends HttpServlet {
+
+    private ArrayList<String> dishes;
```
bprusak: You should first establish a common data model (i.e. how is a dish represented in your Java code) before adding logic like this.
If you use `String` to represent the entire object, you'll need to soon recycle your code to use Olga's `Meal` Java object representing the data instead.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.gson.Gson;
+
+@WebServlet("/search-meal")
```
bprusak: Please move this endpoint from "/search-meal" to another place as per instructions below.

This endpoint is not very RESTful, and REST is a good convention to follow for this project - it fits the usecase for every functionality you need your servlets for.
According to a [Stackoverflow comment](https://softwareengineering.stackexchange.com/questions/233164/how-do-searches-fit-into-a-restful-interface) that I found after comparing multiple approaches proposed, the best solution is to put search at "/meal" endpoint (that represents where you can get resources called "meal" from your server), and this endpoint will be **shared** with the one that just retrieves data for a chosen meal (for example - for when you open the meal's main page), that **Olga put in her code** in the pull request #2 . They both would also be a GET operation in HTTP terms.
The difference will be that for the search you'd need to pass the search params as GET request params, i.e. the URL will look like this: `mealassistant.com/meal?query=tomato%20lasagne`. Meanwhile the request for a single entry might have something like entry ID right in the URL (i.e. `/meal/12345`). You'll need some code that can tell these situations apart, and you'll need to share that place with Olga.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
```diff
@@ -0,0 +1,77 @@
...
+        dishes.add("fried potato with mushrooms and onion");
+        dishes.add("pizza with pineaple, sausage and tomato");
+        dishes.add("vegetable soup with onion");
+        dishes.add("chocolate cake with butter cream and strawberry");
+        dishes.add("hot chocolate with sugar and caramel");
+    }
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String gson = new Gson().toJson(result);
```
bprusak: Please make `result` a local variable in the function(s) you use.

Here you're getting the `result` variable, which might be empty, or might be different depending on what kind of POST request was called beforehand. You have no guarantee which of the two will happen, because the servlet may be restarted in the meantime - especially on AppEngine, where under heavy load multiple instances of this servlet will be started, each running independently, and the user's request may hit a different instance every time.

src/main/webapp/search-meal-script.js
```diff
@@ -0,0 +1,31 @@
...
+        let container = document.getElementById("dishes-container");
+        container.innerText = "";
+        dishes = dishes ?? [""];
+        dishes.forEach((dish) => {
+            container.appendChild(createListElement(dish));
+        });
+    });
+}
+
+function createListElement(val) {
```
bprusak, line 27: Both you and Olga make very similar code, because you both make the very starter basics of your website. After you both submit all you have, take a look to refactor the code and have similar (or "exactly same") functions in some utility "library" that will be common for you two.

src/main/webapp/style.css
```diff
@@ -0,0 +1,13 @@
+/* Copyright 2020 Google LLC
```
bprusak: This file will be just a source of conflicts, as Olga uploads one that has some actual code.
Consider not uploading this file (you'll need to change index.html accordingly for the time being).

src/main/java/com/google/sps/servlets/SearchMealServlet.java
polinasand: Of course I'll change it, I started my part earlier, thats why I used `String`.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
polinasand: So as I understand, we create only one servlet with REST methods?

polinasand: I'm not sure if you mean that, but I did similar here: 
https://github.com/sandatsian/Database-Problems/blob/master/BDProblems/Controllers/ProblemsController.cs 
on C#. Here I have a list of methods for class `Problem` and in each method has `Id` in params, so Url looks like `Problems/Edit/<Id>` when I click Edit button.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
polinasand: Well, I dont exactly understand how I have to keep data about result of search.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
zajonc-gugle: It's usually a good idea to clearly mark temporary solutions in the code. Some kinda comment, e.g.
```
TODO(sandatsian): replace String with Meal
```
would do.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
zajonc-gugle: The suggestion here follows what @bprusak suggested above in the comment regarding RESTful services.

Long story short, if I understand correctly, the suggestion is to use only `doGet` to perform the search. It would require removing the current `doPost` method completely. This way you wouldn't need to store `result` in a private variable, because it would be both created and used in the same method (here, `doGet`).

In general, the less private properties representing some kinda state a class has, the less likely it is to have a bug. Managing state is tricky, and the easiest way of solving the problem is by avoiding it.

Let's take a look at the following example. From what I understand, there's an assumption now that a user performs a `POST` request which populates the `result` variable, and soon after the same user performs a `GET` request, which fetches the stored data.
What would happen if we have 2 users using the app at the same time?
1. User A sends a `POST`, searching for "chocolate". We set `result= ["hot chocolate with sugar and caramel"]` and return a OK response.
2. Soon after that, User B sends a `POST` searching for "potato". We set `result=["fried potato with mushrooms and onion"]`.
3. User A wants to retrieve his search results, expecting some chocolate-related suggestions. We encode `result` and return it to the user. However, in step 2. we set `result` to a new value: french fries, and that's the result User A receives.

If we didn't store the temporary state in a private variable but kept it local (by doing both fetching and returning the result in the same request), no such problems would appear.

src/main/java/com/google/sps/servlets/SearchMealServlet.java
```diff
@@ -0,0 +1,77 @@
...
+
+        response.setContentType("application/json");
+        response.getWriter().println(gson);
+    }
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        result = searchDishByKeyWord(getParameter(request, "text", ""));
```
zajonc-gugle: Note, that this will probably return the whole list of dishes. In the long term we'd probably want to limit the results (kinda like the Google search engine - oftentimes you get millions of matches, but you receive only the top 20; the response would've been far too big otherwise).

src/main/webapp/search-meal-script.js
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+function searchMeal() {
+    fetch('/search-meal').then(response => response.json()).then((dishes) => 
+    {
+        let container = document.getElementById("dishes-container");
```
zajonc-gugle: Let's use `const` instead of `let`.

zajonc-gugle (changes requested):

src/main/java/com/google/sps/servlets/SearchMealServlet.java
polinasand: I will resolve it in the next steps when I will create REST API

src/main/java/com/google/sps/servlets/SearchMealServlet.java
bprusak: I want it to look like this:
* `GET meal/<Id>` gets info of the meal,
* `GET meal/?query=tomato` gets search result for meals where the keyword was "tomato"

If we were to expand upon REST API, it could be extended to:
* `GET meal/` gets a list of meals (i.e. simplification of the search functionality),
* `POST meal/` inserts a new meal (its ID and other details are in POST request body).

`Problems/Edit/<Id>` is  a good URL that describes UI - the hierarchy shows what kind of page should be shown to the user. While you can have URLs like that for managing user-accessible sites, right now we're talking about URLs for data management via API, i.e. for non-human exchange between frontend app (written in Javascript) and the backend app (written in Java). For that we're aiming for some standarized approach (REST) to have some simple URLs (and use various HTTP "verbs": `GET, POST, PUT, DELETE`) to define this software-to-software interface.
So you can have an (admin) user accessing the `/meals-page/edit/<Id>` site, which is just an html+Javascript without any servlet behind it. Then, once all edits are made in the UI and the "send" button is clicked, the Javascript calls `PATCH /meal/<Id>` to actually write the change.

src/main/java/com/google/sps/data/Meal.java
```diff
@@ -0,0 +1,37 @@
...
+        this.title = title;
+        this.description = description;
+        this.ingredients = new ArrayList<String>(ingredients);
+        this.type = type;
+    }
+
+    public long getId() {
+        return this.Id;
+    }
+}
```
grenlayk: As I mentioned in my PR, let's leave my file after merging.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -0,0 +1,124 @@
...
+    // Get meal
+	// GET/meal/
+	// GET/meal/Id
+    @Override
+	public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+		String pathInfo = request.getPathInfo();
+
+		if(pathInfo == null || pathInfo.equals("/")){
+			String gson = new Gson().toJson(dishes);
+            response.setContentType("application/json");
```
grenlayk: Seems like formatting is a bit off.
You can use something like [this](https://www.tutorialspoint.com/online_java_formatter.htm) or another Java onlina formatter.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -0,0 +1,124 @@
...
+*/
+
+    // Get meal
+	// GET/meal/
+	// GET/meal/Id
+    @Override
+	public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+		String pathInfo = request.getPathInfo();
+
+		if(pathInfo == null || pathInfo.equals("/")){
```
grenlayk: Nit:
Probably it will be a good idea to add comments that describe proceeded paths right next to the pieces of code that do this stuff.
For example
```
// GET/meal/ or empty path
if (pathInfo == null || pathInfo.equals("/")) {
...
}
```

grenlayk (approved):

src/main/java/com/google/sps/servlets/MealServlet.java
polinasand: Its weird because in my code editor everything is formatted.

grenlayk: Probably it is because of tab / spaces difference and etc.
But still on github page it's messy :(

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -0,0 +1,124 @@
...
+
+@WebServlet("/meal/*")
+public class MealServlet extends HttpServlet {
+
+    private ArrayList<Meal> dishes;
+
+    @Override
+    public void init() {
+        dishes = new ArrayList<>(Arrays.asList());
+        dishes.add(new Meal(0, "fried potato", "fried potato with mushrooms and onion",
```
grenlayk: @sandatsian I've been asked to delete `int` constructor for `Meal`. So could you please change `MealServlet.java` to be able to work with `Long` type?
```
new Meal(0L, ...)
```
`getId()` also will return `Long`

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -0,0 +1,124 @@
...
+            response.getWriter().println(gson);
+			return;
+		}
+		
+		if(pathInfo.split("/").length != 2) {
+			response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+			return;
+		}
+
+        String Id = pathInfo.replaceAll("/", "");
```
grenlayk: You could update this file with approach Bartosz and Tomasz mentioned during standup.
Just describe cases in doGet() method and implement methods below.
Example:
```
// GET /meal/id
if (...) {
  getMealById();
} 
// GET /meal
if (...) {
  searchMeal();
}
```

src/main/java/com/google/sps/data/Meal.java
bprusak: This branch needs to have `master` merged into it before submitting.
The merge will mean that all the changes Olga submitted will be available from the perspective of this pull request.
In particular this means you'll have Olga's implementation of `Meal` class available and you can use that instead of this temporary implementation.
Unless this file does not collide with Olga's submission at all, then you can make this unification in the next pull request.

src/main/java/com/google/sps/servlets/MealServlet.java
zajonc-gugle: Yep, looks like this file uses tabs.

Whatever editor you're using, it should have an option to automatically replace tabs with spaces. I think this is the main reason people prefer spaces over tabs - tabs may have different size depending on the settings, while there's only one way to render a whitespace.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -0,0 +1,124 @@
...
+import com.google.sps.data.Meal;
+
+@WebServlet("/meal/*")
+public class MealServlet extends HttpServlet {
+
+    private ArrayList<Meal> dishes;
+
+    @Override
+    public void init() {
+        dishes = new ArrayList<>(Arrays.asList());
```
zajonc-gugle: Given that we're going to fetch items by id, I'd suggest storing items in a [HashMap](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html) rather than a list. It's got two advantages:
1. The relation between the key and the value is clear. Whenever you add an item to a hashmap, you have to specify what the key is gonna be.
2. Performance is superior if you need to look an item up by an id. For a hashmap, the complexity is O(1). For a list, it's O(n) (you can see it in your implementation below - to find a single item, you have to iterate the whole list, hence O(n))

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -0,0 +1,124 @@
...
+            new ArrayList<>(Arrays.asList("flour", "water", "sausage", "tomato", "pineaple", "cheese")), "pizza"));
+        dishes.add(new Meal(2, "vegetable soup", "vegetable soup with onion",
+            new ArrayList<>(Arrays.asList("potato", "onion", "cabbage", "mushrooms", "water", "carrot", "pumpkin")), "soup"));
+        dishes.add(new Meal(3, "chocolate cake", "chocolate cake with butter cream and strawberry",
+            new ArrayList<>(Arrays.asList("flour", "water", "butter", "cacao", "chocolate", "sugar", "strawberry", "eggs")), "dessert"));
+        dishes.add(new Meal(4, "hot chocolate", "hot chocolate with sugar and caramel",
+            new ArrayList<>(Arrays.asList("cacao", "sugar", "milk", "caramel", "vanil")), "drinks"));
+    }
+
+    /*@Override
```
zajonc-gugle: Let's aim to remove dead code, rather than commenting it.

src/main/java/com/google/sps/servlets/MealServlet.java
zajonc-gugle: I agree. Having separate, private methods for `getMealById` and `searchMeal` is a good idea.

zajonc-gugle (changes requested):

src/main/java/com/google/sps/servlets/MealServlet.java
polinasand: Then do we still need `id` field in class `Meal`?

zajonc-gugle: That's a good question.
I'd say "no", to avoid having two sources of the same data.

src/main/java/com/google/sps/servlets/MealServlet.java
polinasand: Do you think is it better to remove few commits where I started develop REST API and merge servlet that I had before, like Olga did? And start develop REST in new PR, so we could change same file.

zajonc-gugle: Well, removing commits is somewhat tedious, and we're already in the middle of a review, so I'd probably suggest continuing with what we already have.

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: Please remember - this is all temporary. In the end, instead of effectively searching a structure in memory, you'll need to effectively search a (Datastore) database.
In such case you'll need to put `id` inside the `Meal` entry anyway.
At that point, even using a `List<>` isn't the worst - it's only temporary and only for the list of 5 items below. O(n) won't be bad for the search algorithm for this scale. For the future you'll need to abandon this code and replace it with database queries anyway.

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: I disagree with Tomasz on that one. I think you should grab all this code, put it into a commit in a new branch, and then make a commit deleting all this commented code in this branch.
Step-by-step to make it easy:
1. copy all this commented code to a separate file, outside of the repository - just to be 100% sure it's safe.
2. make a new commit in this branch that deletes this comment.
3. make a new branch, BUT instead of basing it on master, base it on this branch (i.e. first step is `git checkout prototype-polina` instead of `git checkout master`).
4. now you have a branch that assumes code from the previous PR is already submitted, whether it's true or not.
5. in this new branch put all this code in a new commit.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -0,0 +1,124 @@
...
+
+        String Id = pathInfo.replaceAll("/", "");
+
+        List<Meal> meal = dishes
+            .stream()
+            .filter(dish -> String.valueOf(dish.getId()).equals(Id))
+            .collect(Collectors.toList());
+        
+        if (meal.size() > 0) {
+            String gson = new Gson().toJson(meal);
```
bprusak: Please return only a single meal, instead of a Json-ified array of meals. Use `.get(0)` to get the element.
You can put this fix in your follow-up PR that RESTifies your servlet properly.

src/main/java/com/google/sps/servlets/MealServlet.java
zajonc-gugle: ... having said that, it's up to you. Choose whatever's more convenient.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -0,0 +1,124 @@
...
+		if(pathInfo.split("/").length != 2) {
+			response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+			return;
+		}
+
+        String Id = pathInfo.replaceAll("/", "");
+
+        List<Meal> meal = dishes
+            .stream()
+            .filter(dish -> String.valueOf(dish.getId()).equals(Id))
```
bprusak: Please manage the types the other way round - first, before the filter, convert `String` to `Long`, and then in the filter, compare the two `Long`s together.
This makes it better optimized for two reasons: there's only one `String` <=> number conversion to do now, and number-to-number is much cheaper than string-to-string comparison.

bprusak (changes requested):

bprusak (approved):

PR #4, opened 2020-08-17
Update MealServet according to REST

grenlayk:
New servlet instead of old versions. More structured.

### DIFF
--- a/src/main/java/com/google/sps/servlets/MealServlet.java
+++ b/src/main/java/com/google/sps/servlets/MealServlet.java
@@ -1,47 +1,117 @@
 // Copyright 2020 Google LLC
-
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-
+//
 //     https://www.apache.org/licenses/LICENSE-2.0
-
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
 package com.google.sps.servlets;
 
 import java.io.IOException;
-import javax.servlet.annotation.WebServlet;
-import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.stream.Collectors;
+import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.gson.Gson;
 import com.google.sps.data.Meal;
+import java.util.HashMap;
 
-@WebServlet("/meal")
+@WebServlet("/meal/*")
 public class MealServlet extends HttpServlet {
 
-  Meal meal;
+    private final HashMap < Long, Meal > dishes = new HashMap < Long, Meal > ();
 
-  // TODO(grenlayk): implement getting meal info by id (probably in another servlet according to REST API)
+    @Override
+    public void init() {
+        // TODO(sandatsian): implement uploading of dataset from website here
+        dishes.put(0L, new Meal(0L, "fried potato", "fried potato with mushrooms and onion",
+            new ArrayList < > (Arrays.asList("potato", "onion", "mushrooms", "oil")), "Main"));
+        dishes.put(1L, new Meal(1L, "Italian pizza", "Pizza with pineaple, sausage and tomato",
+            new ArrayList < > (Arrays.asList("flour", "water", "sausage", "tomato", "pineaple", "cheese")), "pizza"));
+        dishes.put(2L, new Meal(2L, "vegetable soup", "vegetable soup with onion",
+            new ArrayList < > (Arrays.asList("potato", "onion", "cabbage", "mushrooms", "water", "carrot", "pumpkin")), "soup"));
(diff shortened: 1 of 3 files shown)


polinasand (approved):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -1,47 +1,117 @@
...
 import com.google.gson.Gson;
 import com.google.sps.data.Meal;
+import java.util.HashMap;
 
-@WebServlet("/meal")
+@WebServlet("/meal/*")
 public class MealServlet extends HttpServlet {
 
-  Meal meal;
+    private final HashMap < Long, Meal > dishes = new HashMap < Long, Meal > ();
```
bprusak: Please do *not* put spaces around the `<>` braces and around the words inside the braces. Please replace with `private final HashMap<Long, Meal> dishes`.

This also applies to empty `<>` braces in the `init()` method below - they should be  
`new ArrayList<>(Arrays.asList(...))`.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -1,47 +1,117 @@
...
+        String pathInfo = request.getPathInfo();
+
+        // Get meal
+        // GET/meal/
+        if (pathInfo == null || pathInfo.equals("/")) {
+            searchMeal(request, response);
+            return;
+        }
+
+        // GET/meal/*/*
```
bprusak: Add a comment that this format is invalid. With that comment it's easy to figure out that the following code is about returning an HTTP error.
Also it's easier to understand when this option is put as the last one in the code.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -1,47 +1,117 @@
...
+        dishes.put(4L, new Meal(4L, "hot chocolate", "hot chocolate with sugar and caramel",
+            new ArrayList < > (Arrays.asList("cacao", "sugar", "milk", "caramel", "vanil")), "drinks"));
+    }
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String pathInfo = request.getPathInfo();
+
+        // Get meal
+        // GET/meal/
```
bprusak: Replace `GET/meal/` with `GET meal/`. The verb is not a part of the URL.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -1,47 +1,117 @@
...
+            getMealById(request, response);
+            return;
+        }
+
+        response.sendError(HttpServletResponse.SC_NOT_FOUND);
+        return;
+    }
+
+    private void searchMeal(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String gson = new Gson().toJson(dishes);
```
bprusak, line 83: Add a `TODO(grenlayk)` comment that says this is just a hardcoded result, and that you'll replace it with actual search in the future. Also specify whether you'll add search for prototype or only later, for MVP.

bprusak (changes requested):

grenlayk:
I merged master into this branch to be able to to continue work in another branch.
(Which I created from this one.)

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -1,47 +1,120 @@
...
 // limitations under the License.
-
 package com.google.sps.servlets;
 
 import java.io.IOException;
-import javax.servlet.annotation.WebServlet;
-import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Arrays;
```
bprusak, line 18: Please, try to keep alphabetical order of imports.
I see that you put `javax.servlet.annotation.WebServlet` in its rightful place, but now `java.util.Arrays` should be moved back up to the second place, before `ArrayList`.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -1,47 +1,120 @@
...
 import com.google.gson.Gson;
 import com.google.sps.data.Meal;
+import java.util.HashMap;
 
-@WebServlet("/meal")
+@WebServlet("/meal/*")
 public class MealServlet extends HttpServlet {
 
-  Meal meal;
+    private final HashMap <Long, Meal> dishes = new HashMap <Long, Meal> ();
```
bprusak: Please remove the spaces around the `<>` brackets if not needed, i.e. instead of:  
`private final HashMap <Long, Meal> dishes`, please do:
`private final HashMap<Long, Meal> dishes`.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -1,47 +1,120 @@
...
+
+    @Override
+    public void init() {
+        // TODO(sandatsian): implement uploading of dataset from website here
+        dishes.put(0L, new Meal(0L, "fried potato", "fried potato with mushrooms and onion",
+            new ArrayList <> (Arrays.asList("potato", "onion", "mushrooms", "oil")), "Main"));
```
bprusak: Please remove the extra spaces around `<>` brackets, that is, instead of:  
`new ArrayList <> (Arrays.asList(...))`, please do:  
`new ArrayList<>(Arrays.asList(...))`.

bprusak (approved):

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: Actually I'm wrong, both as per alphabetic order AND as per [style guide](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing) demanding ASCII order, the `ArrayList` comes just before `Arrays`, because capital `L` is before lowercase `s`. So it should be: `ArrayList, Arrays, List`.

PR #6, opened 2020-08-18
Add restaurants and locate user

grenlayk:
I know that part with `addRestaurants()` is awful, but haven't found how not to use it 3 times, but still first locate use and only  after that add restaurants' markers.

### DIFF
--- a/src/main/webapp/meal-page-script.js
+++ b/src/main/webapp/meal-page-script.js
@@ -13,24 +13,27 @@
 // limitations under the License.
 
 function fetchMealInfo() {
-    fetch('/meal').then(response => response.json()).then((meal) => 
-    {
-        const { title, description, ingredients } = meal;
-        const titleElement = document.getElementById("title");
-        titleElement.innerText = encodingCheck(title);
-        const descriptionElement = document.getElementById("description");
-        descriptionElement.innerText = encodingCheck(description);
-        const ingredientsElement = document.getElementById("ingredients");
-        for (const ingredient of ingredients) {
-            ingredientsElement.appendChild(createElementByTag(encodingCheck(ingredient), 'li'));
-        }
-        createMap();
+  fetch("/meal/0")
+    .then((response) => response.json())
+    .then((meal) => {
+      const { title, description, ingredients } = meal;
+      const titleElement = document.getElementById("title");
+      titleElement.innerText = encodingCheck(title);
+      const descriptionElement = document.getElementById("description");
+      descriptionElement.innerText = encodingCheck(description);
+      const ingredientsElement = document.getElementById("ingredients");
+      for (const ingredient of ingredients) {
+        ingredientsElement.appendChild(
+          createElementByTag(encodingCheck(ingredient), "li")
+        );
+      }
+      createMap();
     });
 }
 
 // solution from https://stackoverflow.com/questions/20174280/nodejs-convert-string-into-utf-8
 function encodingCheck(string) {
-    return JSON.parse(JSON.stringify(string));
+  return JSON.parse(JSON.stringify(string));
 }
 
 function createElementByTag(text, tag) {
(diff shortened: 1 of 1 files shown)


polinasand (approved):

src/main/webapp/meal-page-script.js
```diff
@@ -39,21 +42,79 @@ function createElementByTag(text, tag) {
   return element;
 }
 
-/** Creates a map and adds it to the page. */
+let map;
```
zajonc-gugle: If I understand correctly, this is a global variable (defined outside of a function).
Global variables are quite evil. It's better to avoid them whenever possible. It's only used by `createMap` and `addRestaurants`, but `addRestaurants` is called from inside `createMap`.
I'd suggest:
1. not declaring the variable here
2. declaring it in `createMap`
3. passing it as an argument to `addRestaurants`.

src/main/webapp/meal-page-script.js
```diff
@@ -39,21 +42,79 @@ function createElementByTag(text, tag) {
...
+
 function createMap() {
-  const userLocation = new google.maps.LatLng(55.746514, 37.627022);
+  let userLocation = new google.maps.LatLng(55.746514, 37.627022);
```
zajonc-gugle: `const` is preferred over `let`.

src/main/webapp/meal-page-script.js
```diff
@@ -39,21 +42,79 @@ function createElementByTag(text, tag) {
...
+    const marker = new google.maps.Marker({
+        position: restaurantLocation,
+        map,
+        title: "Restaurant",
+        animation: google.maps.Animation.DROP
+    });
+  }
+}
+
+function changeLocation(location, latDiff, lngDiff) {
```
zajonc-gugle: Let's use a more descriptive name, e.g. `moveLocationBy`

src/main/webapp/meal-page-script.js
```diff
@@ -39,21 +42,79 @@ function createElementByTag(text, tag) {
...
+    const lat = location.lat() + latDiff;
+    const lng = location.lng() + lngDiff;
+    return {
+        lat, 
+        lng
+    };
+}
+
+function handleLocationError(browserHasGeolocation) {
```
zajonc-gugle: I'd suggest avoiding functions with a single, `boolean` argument that only have an `if..else` block inside.
We can have two separate functions instead.

src/main/webapp/meal-page-script.js
```diff
@@ -39,21 +42,79 @@ function createElementByTag(text, tag) {
...
+  } else {
+    handleLocationError(false);
+    addRestaurants();
+  }
+}
+
+function addRestaurants() {
+  // Hard coded restaurants
+  const restaurants = [];
+  restaurants.push(changeLocation(map.getCenter(), 0.001, 0.001));
```
zajonc-gugle: Nit:
this is a Java-ish approach: define an empty list and then call `push` a few times.
In JS you'd normally do it in one line:
```
const restaurants = [changeLocation(map.getCenter(), 0.001, 0.001), changeLocation(map.getCenter(), -0.001, 0.001)];
```

zajonc-gugle (changes requested):

src/main/webapp/meal-page-script.js
```diff
@@ -39,21 +42,79 @@ function createElementByTag(text, tag) {
...
     position: userLocation,
     map,
-    title: "You are here",
+    title: "You are here (probably)",
+    label: "U",
     animation: google.maps.Animation.DROP,
   });
+
+  
+  if (navigator.geolocation) {
```
zajonc-gugle: Oh, so roughly speaking, we have a function with a rather unfriendly API (`getCurrentPosition`). One common way to work with such functions is to wrap them in another custom function with a more friendly API.

This callback-style code was common before introducing Promises [some more details here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises). It often leads to the problem of having to pass the same callback multiple times and being quite hard to follow (aka the callback hell).

The good news is that all of those callback-style functions can be rewritten in terms of promises. The `Promise` object has a constructor of
```
const p = new Promise((resolve, reject) => {
  // some asynchronous code here. The key bit is that `p` will be a Promise, so you can hide the oldschool api here and never think about it again
});
```

If you look at the constructor signature, it's:
1. a function (like every constructor)
2. that takes a function as an argument (the argument passed to `Promise(...)` has a `=>`, therefore it's a function
3. which takes two functions as its argument (both `resolve` and `reject` are functions)
I understand that this may seem a bit weird, so I'm not gonna bore you with the details, but this kinda functional programming (here functional meaning - treating functions just like any other objects, e.g. passing them around as arguments, returning them as values, etc.) is quite common in JS, but can be minimized if someone doesn't feel comfortable working with them.
Btw. those functions that operate on functions are often called [Higher Order Functions](https://en.wikipedia.org/wiki/Higher-order_function) (or HOFs for short).

Either way, you can hide the oldschool geolocation object behind a modern API with this magic function:
```
function getCurrentPositionPromise() {
  return new Promise((resolve, reject) => {
    if (!navigator.geolocation) {
      reject("NO_GEOLOCATION");  // Rejections can be read with a `Promise.catch` method.
    } else {
      const onSuccess = (position) => resolve(position);  // This value will be available on the first `Promise.then` call.
      const onError = () => reject("GET_POSITION_FAILED");
      navigator.geolocation.getCurrentPosition(onSuccess, onError);
    }
  });
}
```
and use it like any other promise, e.g.
```
getCurrentPositionPromise().then(pos => console.log("Current position is:", pos)).catch(err => console.log("this bit will only be called if the promise was rejected", err)).then(() => console.log("another callback"));
```

If you want to experiment with it, you can just paste the function into your browser's console (F12).

grenlayk: Thanks!

bprusak (approved):

zajonc-gugle (approved):

PR #7, opened 2020-08-18
Meal HTML pages and search page

polinasand:
Create search view (without searching actually), meal.html view for one hard coded object `Meal` (index 0)

### DIFF
--- a/src/main/webapp/meal-page-script.js
+++ b/src/main/webapp/meal-page-script.js
@@ -13,7 +13,13 @@
 // limitations under the License.
 
 function fetchMealInfo() {
-    fetch('/meal/0').then(response => response.json()).then((meal) => 
+    // use mapping /meal.html?id=<id> 
+    // fetches form server by action meal/<id>
+    const queryString = window.location.search;
+    const urlParams = new URLSearchParams(queryString);
+    const id = urlParams.get("id");
+
+    fetch('/meal/' + id.toString()).then(response => response.json()).then((meal) => 
     {
         const { title, description, ingredients } = meal;
         const titleElement = document.getElementById("title");
--- a/src/main/webapp/meal-page-script.js
+++ b/src/main/webapp/meal-page-script.js
@@ -37,6 +37,7 @@ function fetchMealInfo() {
         }
         createMap();
     }).catch((error) => {
+        //TODO(sandatsian): display error page
         console.log(error);
     });
 }
--- a/src/main/webapp/meal-page-script.js
+++ b/src/main/webapp/meal-page-script.js
@@ -15,7 +15,6 @@
 function fetchMealInfo() {
     fetch('/meal/0').then(response => response.json()).then((meal) => 
     {
-        console.log(meal);
         const { title, description, ingredients } = meal;
         const titleElement = document.getElementById("title");
         titleElement.innerText = encodingCheck(title);
--- a/src/main/webapp/meal-page-script.js
+++ b/src/main/webapp/meal-page-script.js
@@ -17,9 +17,14 @@ function fetchMealInfo() {
     // fetches form server by action meal/<id>
     const queryString = window.location.search;
     const urlParams = new URLSearchParams(queryString);
-    const id = urlParams.get("id");
+    const id = urlParams.get("id") ?? 0;
 
-    fetch('/meal/' + id.toString()).then(response => response.json()).then((meal) => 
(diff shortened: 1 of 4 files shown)


src/main/webapp/meal-page-script.js
```diff
@@ -13,7 +13,13 @@
 // limitations under the License.
 
 function fetchMealInfo() {
-    fetch('/meal').then(response => response.json()).then((meal) => 
+    // use mapping /meal.html?id=<id> 
+    // fetches form server by action meal/<id>
+    const queryString = window.location.search;
+    const urlParams = new URLSearchParams(queryString);
+    const id = urlParams.get("id");
```
grenlayk: Probably a good idea to check `null` case. Make a request to default value for example.

grenlayk (approved):

polinasand:
Update: Create `index.html` for searching meal, where actually there is only list of all meals (for now).
For each list component made a link, that redirects to one `meal.html` page (url `meal.html?id=<id>`).
Then get an `id` from url path, fetching data from server using action `meal/<id>`. (`meal-page-script.js`).

src/main/webapp/meal-page-script.js
```diff
@@ -13,7 +13,13 @@
...
+
+    fetch('/meal/' + id.toString()).then(response => response.json()).then((meal) =>
```
bprusak: Once you get a response, you should check whether `response.ok` is true.

If the backend responds with some kind of error (such as "not found"), then the response will still be there. However, the status of the response will be set to an error code (usually 4XX). To check that quickly, the `Response` interface has an `ok` property that is true only if the status code points to success (that is: 2XX).

When `fetch` gets a "404 not found" response, `.json()` may or may not succeed - there may be a body to the error response (such as an error message), and in both cases the consequences may be confusing. To avoid that, please check the response status.

src/main/webapp/search-meal-script.js
```diff
@@ -17,15 +17,20 @@ function searchMeal() {
...
-            container.appendChild(createListElement(dish));
+        dishes = dishes ?? {0: ""};
+        Object.entries(dishes).forEach((dish) => {
+            container.appendChild(createListElement(dish[1]));
         });
     });
 }
 
-function createListElement(val) {
+function createListElement(obj) {
```
bprusak: You can name the argument "dish" so that it's easier to assume what kind of object is this.

bprusak (changes requested):

src/main/webapp/search-meal-script.js
```diff
@@ -17,15 +17,20 @@ function searchMeal() {
...
     const listElement = document.createElement('li');
-    listElement.innerText = val;
+    const id = obj.id;
+    listElement.innerText = obj.title;
+    listElement.onclick = function() {
+        location.href = 'meal.html?id=' + id.toString();
```
zajonc-gugle: There's a special syntax for building strings in JS: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

```
`meal.html?id=${id}`
```

src/main/webapp/meal-page-script.js
zajonc-gugle: Yep.
One thing that may be a bit counterintuitive in the fetch api is that it will only reject the promise on a network error, not a http error [link](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch). Another words, the `fetch` promise will get rejected if you lose your internet connection, but not if you point it to a non-existent url.
Handling non-ok statuses has to be done manually by inspecting the `ok` field. You can find some examples on the MDN page, or you can google something along the lines of "fetch http error handling".

zajonc-gugle (changes requested):

src/main/webapp/meal-page-script.js
```diff
@@ -25,6 +36,8 @@ function fetchMealInfo() {
             ingredientsElement.appendChild(createElementByTag(encodingCheck(ingredient), 'li'));
         }
         createMap();
+    }).catch((error) => {
+        console.log(error);
```
bprusak: Add a TODO for the Alpha product: if the endpoint returned "not found", display an error on the page explaining what happenned. For other types of errors, display a generic "something went wrong" error page.

bprusak (approved):

PR #8, opened 2020-08-19
Find similar

grenlayk:


polinasand (approved):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -111,6 +119,18 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
         return;
     }
 
+     private void returnUrlToSimilar(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        // Fuction redirect to random page
+        // TODO(grenlayk): implement suggestions algorithm here for Product Alpha
+        Random rand = new Random(); 
+        int randomId = rand.nextInt(dishes.size()); 
+        String url = "/meal.html" + "?id=" + Integer.toString(randomId);
```
zajonc-gugle: I believe you can use `StringBuilder` here, instead of concatenating string.

zajonc-gugle: I'd suggest letting the frontend build a full url. It should be sufficient for the backend to just return the Id of a suggested meal.

zajonc-gugle (changes requested):

src/main/webapp/meal-page-script.js
```diff
@@ -55,7 +55,14 @@ function createElementByTag(text, tag) {
   return element;
 }
 
+function redirectToSimilar() {
+    fetch('/meal/similar').then(response => response.json()).then((id) => {
+        const url = '/meal.html?id=' + id.toString();
```
zajonc-gugle: You can use a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -111,6 +119,17 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
         return;
     }
 
+     private void returnUrlToSimilar(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
zajonc-gugle: You can rename the method now. It no longer returns a url.

zajonc-gugle (approved):

PR #12, opened 2020-08-21
Datastore in Meal Servlet

polinasand:


src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -14,11 +14,22 @@
...
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Query.SortDirection;
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.ArrayList;
+import java.util.Arrays;
```
grenlayk, line 30: Seems like it is better to order libraries in lexicographic order.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -33,21 +44,47 @@
...
+            new Meal(7L, "Chocolate Chips Pancake", "So delicious and chocolate breakfast.",
+                new ArrayList<>(Arrays.asList("2 eggs", "salt", "butter", "flour", "sugar", "chocolate chips", "milk")), "Pancake")));
+        datastore.put(createMealEntity(
+            new Meal(8L, "Seafood Pizza", "Just from the bottom of the ocean...",
+                new ArrayList<>(Arrays.asList("flour", "salt", "butter", "water", "oil", "shrimp", "tuna", "onion", "ctabstick",
+                "scallops", "fish sticks", "garlic", "tomato sauce")), "Pizza")));
+        datastore.put(createMealEntity(
+            new Meal(9L, "Almond Fudge Brownies", "There is any almond lover's prime recipe for fudge brownies.",
+                new ArrayList<>(Arrays.asList("2 eggs", "sugar", "butter", "milk", "almond", "vanilla", "cocoa powder")), "Dessert")));
+        // TODO: create 5 new ones
```
grenlayk: Not sure that this is needed. Probably it's better to start learning how to load everything from wiki.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -57,16 +94,17 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
         // GET meal/
         if (pathInfo == null || pathInfo.equals("/")) {
-            searchMeal(request, response);
+            getMealList(request, response);
             return;
         }
 
-        
         if (pathInfo.split("/").length == 2) {
             String idString = pathInfo.replaceAll("/", "");
```
grenlayk: I know I named it like that myself first time, but now seems like name doesn't match variable anymore.
Maybe `requestType` or `requestInfo` will be better.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -85,13 +123,36 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+        mealEntity.setProperty("title", meal.getTitle());
+        mealEntity.setProperty("description", meal.getDescription());
+        mealEntity.setProperty("ingredients", meal.getIngredients());
+        mealEntity.setProperty("type", meal.getType());
+ 
+        return mealEntity;
+    }
+
+    private void getMealList(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        Query query = new Query("Meal").addSort("id", SortDirection.DESCENDING);
```
grenlayk: I suppose it is better to sort entities by `title`, it is more common for user.
Or first by `type`, then by `title`.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -85,13 +123,36 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+        response.setContentType("application/json;");
+        response.getWriter().println(gson.toJson(meals));
+        return;
+    }
+
     private void searchMeal(HttpServletRequest request, HttpServletResponse response) throws IOException {
         // Fuction returns hardcoded result
         // TODO(sandatsian): implement search algorithm here for MVP
-        String gson = new Gson().toJson(dishes);
+        /*String gson = new Gson().toJson(dishes);
```
grenlayk: Probably you do not need to comment anything, just delete - you always can find it in previous commits.

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: This order that Polina introduces here is lexicographic: L comes before S.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -33,21 +44,47 @@
...
+        datastore.put(createMealEntity(
+            new Meal(7L, "Chocolate Chips Pancake", "So delicious and chocolate breakfast.",
+                new ArrayList<>(Arrays.asList("2 eggs", "salt", "butter", "flour", "sugar", "chocolate chips", "milk")), "Pancake")));
+        datastore.put(createMealEntity(
+            new Meal(8L, "Seafood Pizza", "Just from the bottom of the ocean...",
+                new ArrayList<>(Arrays.asList("flour", "salt", "butter", "water", "oil", "shrimp", "tuna", "onion", "ctabstick",
+                "scallops", "fish sticks", "garlic", "tomato sauce")), "Pizza")));
+        datastore.put(createMealEntity(
+            new Meal(9L, "Almond Fudge Brownies", "There is any almond lover's prime recipe for fudge brownies.",
+                new ArrayList<>(Arrays.asList("2 eggs", "sugar", "butter", "milk", "almond", "vanilla", "cocoa powder")), "Dessert")));
```
bprusak: Please do not add entries to Datastore during startup.

Datastore is persistent storage that exists independently of your Java application. Once something is written to a Datastore table, it will still be there when you restart / update your app.
So don't add the hard-coded entries here. Instead use Cloud Console interface to manually add entries to your Datastore in `polinasand-step-2020` project.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -57,16 +94,17 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         // Get meal
         // GET meal/
         if (pathInfo == null || pathInfo.equals("/")) {
-            searchMeal(request, response);
+            getMealList(request, response);
```
bprusak, line 53: I don't think the search and the meal listing should be separated.

Both functionalities will list meals for the user. The difference is that search will return only some of the meals, while a regular GET returns all of them.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,9 +166,12 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
         }
-
-        Meal meal = dishes.get(id);
-
+        Filter propertyFilter = new FilterPredicate("id", FilterOperator.EQUAL, id);
+        Query query = new Query("Meal").setFilter(propertyFilter);
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery results = datastore.prepare(query);
+        ArrayList<Meal> meals = getDataFromDatastore(results); 
+        Meal meal = meals.get(0);
```
bprusak: Be prepared for a case where number of meals returned by Datastore is not equal to 1.
If the number is 0, then a "404 NOT FOUND" error should be prepared.
If the number is higher than 1, then something went very wrong - you may want to abort and prepare a "500 INTERNAL SERVER ERROR".

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -33,21 +44,47 @@
...
-            new ArrayList<>(Arrays.asList("Potato", "Onion", "Mushrooms", "Oil")), "main"));
-        dishes.put(1L, new Meal(1L, "Italian pizza", "Pizza with pineaple, sausage and tomato.",
-            new ArrayList<>(Arrays.asList("Flour", "Water", "Sausage", "Tomato", "Pineaple", "Cheese")), "pizza"));
-        dishes.put(2L, new Meal(2L, "Vegetable soup", "Vegetable soup with onion.",
-            new ArrayList<>(Arrays.asList("Potato", "Onion", "Cabbage", "Mushrooms", "Water", "Carrot", "Pumpkin")), "soup"));
-        dishes.put(3L, new Meal(3L, "Chocolate cake", "Chocolate cake with butter cream and strawberry.",
-            new ArrayList<>(Arrays.asList("Flour", "Water", "Butter", "Cacao", "Chocolate", "Sugar", "Strawberry", "Eggs")), "dessert"));
-        dishes.put(4L, new Meal(4L, "Hot chocolate", "Hot chocolate with sugar and caramel.",
-            new ArrayList<>(Arrays.asList("Cacao", "Sugar", "Milk", "Caramel", "Vanil")), "drinks"));
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
zajonc-gugle: I'm wondering whether this code will keep on adding new items whenever we start the server. I'm not sure how this class works - does it create a new DataStore every time we use it, or does it connect to the same instance somewhere?

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -119,15 +183,30 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
         // Fuction redirect to random page
         // TODO(grenlayk): implement suggestions algorithm here for Product Alpha
-        Random rand = new Random(); 
+        /*Random rand = new Random(); 
         int randomId = rand.nextInt(dishes.size()); 
         
         String gson = new Gson().toJson(randomId);
         response.setContentType("application/json");
-        response.getWriter().println(gson);
+        response.getWriter().println(gson);*/
```
bprusak: As Olga mentioned earlier - do not comment old code that is no longer needed, just delete it instead. We can retrieve the old code by browsing the commit history.

src/main/java/com/google/sps/servlets/MealServlet.java
zajonc-gugle: Yeah, I agree. Adding more items manually will probably not enhance the experience too much.

Btw. manually creating data for testing doesn't scale well. A few items are fine, because you can somewhat control their content. However, when one needs to create a larger amount of items, the exact content doesn't matter that much, so it's often easier to just fill the instances with random data.

A classic example of a random text data generator is "lorem ipsum" (there's hundreds of implementations available).
See here for an example of a Java lorem ipsum generator: https://github.com/mdeanda/lorem

Having said that, I think we're about to reach the point where we have to start using some external source of data anyway, so let's not spend too much time adding more hardcoded items.

src/main/java/com/google/sps/servlets/MealServlet.java
zajonc-gugle: In general that's true, but let's not forget that this is not exactly what the user will see.
This is an API method, which is kind expected to be used by advanced users (scripts or developers, basically). It may make sense to return the data in a format that makes sense to a developer, rather than a regular user.

Thinking about the end user can be implemented by the frontend code. The JS script can simply sort the results while rendering.

Or, to put it differently - if we are convinced that sorting items by name makes sense to the end user, we should make the frontend code sort meals, regardless of whether the backend already sorts them or not. If the frontend is sure that this is the right approach, it should just do it, rather than depend on the backend.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -85,13 +123,36 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
         // TODO(sandatsian): implement search algorithm here for MVP
-        String gson = new Gson().toJson(dishes);
+        /*String gson = new Gson().toJson(dishes);
         response.setContentType("application/json");
         response.getWriter().println(gson);
-        return;
+        return;*/
     }
 
     private void getMealById(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
bprusak: I need you to write unit tests for these features - getting a single element by ID and listing all the elements.

In these tests you need to create an instance of the MealServlet class, and then either call the `doGet()` method in the test, or expose the `getMealById` and `getMealList` methods so you can call them in the test.

Since your tests aren't supposed to access the real Datastore in GCE, you'll need to follow this [Local unit testing for Java 8](https://cloud.google.com/appengine/docs/standard/java/tools/localunittesting) AppEngine tutorial. Each test will need to prepare the database for the tested scenario (by configuring the service with `LocalDatastoreServiceTestConfig`, and then adding some hardcoded examples as a part of test's code).

bprusak (changes requested):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -119,15 +183,30 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+            result.add(meal);
+        }
+        return result;
+    }
+
+    private void returnIdOfSimilar(HttpServletRequest request, HttpServletResponse response) throws IOException {
         // Fuction redirect to random page
         // TODO(grenlayk): implement suggestions algorithm here for Product Alpha
-        Random rand = new Random(); 
+        /*Random rand = new Random();
```
zajonc-gugle: I guess those are leftovers from development. I assume you didn't really mean to comment out this code.

zajonc-gugle (changes requested):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -33,40 +44,26 @@
...
-            new ArrayList<>(Arrays.asList("Flour", "Water", "Sausage", "Tomato", "Pineaple", "Cheese")), "pizza"));
-        dishes.put(2L, new Meal(2L, "Vegetable soup", "Vegetable soup with onion.",
-            new ArrayList<>(Arrays.asList("Potato", "Onion", "Cabbage", "Mushrooms", "Water", "Carrot", "Pumpkin")), "soup"));
-        dishes.put(3L, new Meal(3L, "Chocolate cake", "Chocolate cake with butter cream and strawberry.",
-            new ArrayList<>(Arrays.asList("Flour", "Water", "Butter", "Cacao", "Chocolate", "Sugar", "Strawberry", "Eggs")), "dessert"));
-        dishes.put(4L, new Meal(4L, "Hot chocolate", "Hot chocolate with sugar and caramel.",
-            new ArrayList<>(Arrays.asList("Cacao", "Sugar", "Milk", "Caramel", "Vanil")), "drinks"));
-    }
-
+    // TODO(sandatsian): implement uploading of dataset from website
```
bprusak: This uploading should be done separately from this application. Therefore, this TODO does not belong here.

Since there's no file to put that TODO in, make a new Github Issue instead - to upload a dataset to Datastore. (It will be a big task).

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -33,40 +44,26 @@
...
         if (pathInfo.split("/").length == 2) {
-            String idString = pathInfo.replaceAll("/", "");
-            if (idString.equals("similar")) {
+            String requestType = pathInfo.replaceAll("/", "");
+            if (requestType.equals("similar")) {
                 // GET meal/similar
                 returnIdOfSimilar(request, response);
+            } else if (requestType.contains("search")) {
+                // GET meal/search
+                getMealList(request, response);
```
bprusak: Remove this entire `if` condition and its body.

By not separating "search" from "getMealList", I not only meant both to live under `getMealList` function, but also that both should be under `meal/` endpoint. For regular list, the call will have no extra parameters. For a search call, the call will have get params given in the URL, such as `meal/?query=pork`. Servlet's path info should still be `meal/`, but the request params should have a "query" param, with the value set to "pork".

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -85,12 +82,26 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         return;
     }
 
-    private void searchMeal(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        // Fuction returns hardcoded result
-        // TODO(sandatsian): implement search algorithm here for MVP
-        String gson = new Gson().toJson(dishes);
-        response.setContentType("application/json");
-        response.getWriter().println(gson);
+    private Entity createMealEntity(Meal meal) {
```
bprusak, line 84: Add documentation of what this method does, using the "Javadoc" notation.
Either read tutorials on how it's done, **or** open this file in Intellij Idea, and start a comment right above this method by typing `/**` and then enter. Intellij should automatically put the parts you need for the complete documentation (a free spot to describe the method in general, an "arguments" section and "returns" section).

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: `DESCENDING` order will sort the elements starting with the largest number, and then continuing until the smallest number (for example start with 10 and end with 1).
If you wanted to sort the other way (for example from 1 to 10), you should use `ASCENDING`.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,29 +116,56 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+        }
+        if (meals.size() == 1) {
+            Meal meal = meals.get(0);
+            if (meal != null) {
+                String gson = new Gson().toJson(meal);
+                response.setContentType("application/json");
+                response.getWriter().println(gson);
+                return;
+            }
         }
```
bprusak, line 135: You should add a comment saying `meals.size() > 1`, just for readability.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,29 +116,56 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
-            response.getWriter().println(gson);
-            return;
+        Filter propertyFilter = new FilterPredicate("id", FilterOperator.EQUAL, id);
+        Query query = new Query("Meal").setFilter(propertyFilter);
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery results = datastore.prepare(query);
+        ArrayList<Meal> meals = getDataFromDatastore(results); 
+        
+        if (meals.size() > 1) {
+            response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
```
bprusak, line 124: Body of this function is missing a `return` statement, please add it, otherwise the block setting `SC_NOT_FOUND` will also apply.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,29 +116,56 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
 
-     private void returnIdOfSimilar(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    private ArrayList<Meal> getDataFromDatastore(PreparedQuery query) {
+        ArrayList<Meal> result = new ArrayList<>();
+        for (Entity entity : query.asIterable()) {
+            Long id = (Long)entity.getProperty("id");
+            String title = (String) entity.getProperty("title");
+            String description = (String) entity.getProperty("description");
+            ArrayList<String> ingredients = (ArrayList<String>) entity.getProperty("ingredients");
+            String type = (String) entity.getProperty("type");
```
bprusak: You need to put all this in a `try {} catch` block, in case a `ClassCastException` happens (because a field may not convert to the type you expect).
You also need to test what happens if one of these fields is not present, or is empty (such as type="", ingredients being an empty list, etc).

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,29 +116,56 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+        }
+        return result;
+    }
+
+    private void returnIdOfSimilar(HttpServletRequest request, HttpServletResponse response) throws IOException {
         // Fuction redirect to random page
         // TODO(grenlayk): implement suggestions algorithm here for Product Alpha
-        Random rand = new Random(); 
-        int randomId = rand.nextInt(dishes.size()); 
+        Query query = new Query("Meal").addSort("id", SortDirection.DESCENDING);
```
bprusak: Since you're retrieving an item at random, sorting is not needed - please remove it.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,29 +116,56 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery results = datastore.prepare(query);
         
-        String gson = new Gson().toJson(randomId);
-        response.setContentType("application/json");
-        response.getWriter().println(gson);
+        Random rand = new Random(); 
+        int randomId = rand.nextInt(getDataFromDatastore(results).size());
```
bprusak: Id is not necessarily in the range from 0 to "number of entries in database - 1" (i.e. N-1).

For a real-life database, you have to assume that some elements will be missing from the id sequence - usually because a dataset has a long history, and - for example - some elements had to be manually deleted, and in such case no one will redo all the ids. You can't assume what the ids are.

In this case, once you pick a random number X from 0 to N-1, you should read Xth element from the results and get its actual ID - then you'll be sure you have a valid ID.

Also please create a test for this behavior.

bprusak (changes requested):
This code starts to do some transformations on data (mostly - changing the data from Datastore Entities to `Meal` objects). Therefore please add tests for all that functionality.
You should add the tests in the same Pull Request that you introduce a functionality. This time I'll let that pass and you can add tests in a follow-up PR. But later on I will insist on adding CL on same PR. If that would be too big of a change - you'd need to split it up into smaller sub-features (here for example - Meal list using Datastore, single meal using Datastore, similar meal using Datastore), each with its own tests.

src/main/java/com/google/sps/servlets/MealServlet.java
polinasand: In "actual ID" you mean id of `Meal` instance (property this.id) or key of entity?

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: Fill the documentation: after `@param meal` write a sentence describing what this parameter should be.
Similarily, describe what is returned after the `@return` tag.
Nitpick: in every comment (documentation or not), please write them as proper sentences: start with capital letter (i.e. "Create entity (...)" instead of "create entity (...)"), end with a full stop (i.e. "(...) of class Meal." instead of "(...) of class Meal").
https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html has a "Format of a Doc Comment" section that has a nice example of what a complete Javadoc comment looks like.

pom.xml
```diff
@@ -30,14 +30,60 @@
...
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-mock</artifactId>
+        <version>2.0.8</version>
+        <scope>test</scope>
+    </dependency>
+    <dependency>
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-core</artifactId>
+        <version>2.0.8</version>
+        <scope>core</scope>
```
bprusak, line 84: Remove this `spring-core` dependency.

In `spring-mock`'s package definition `spring-core` is listed as a dependency. Therefore, requiring `spring-mock` should automatically download `spring-core` in the required version too.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,29 +115,67 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+        Long ind = 0L;
+        for (Entity entity : query.asIterable()) {
+            try {
+                Long id = (Long)entity.getProperty("id");
+                String title = (String) entity.getProperty("title");
+                String description = (String) entity.getProperty("description");
+                ArrayList<String> ingredients = (ArrayList<String>) entity.getProperty("ingredients");
+                String type = (String) entity.getProperty("type");
+                Meal meal = new Meal(id, title, description, ingredients, type);
+                result.put(ind, meal);
```
bprusak: The key that you use for the `result` Map was supposed to be the `id` that was written in "meal.id", and not just a number from 0 to N.

Also, things changed a bit - you put data into a map, because then you could, for example, find a particular meal by its `id` in an instant.
Now that use case is no longer there, since you're using Datastore's functionality to do that instead.

If you're using `ind` to keep the order for the final list, please abandon returning a `Map` and return a `List` (such as `ArrayList`) insted.

You can keep it a map, but only because the frontend already assumes the result is a map. If it's possible, please change this back into a `List` in the near future.
One of the drawbacks in this situation for using a map is - if you choose to `.addSort` in your query, then it doesn't matter - putting all elements into a HashMap may shuffle the ordering again, and you'll use that order to send the result list to the user.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,140 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import com.google.sps.servlets.MealServlet;
+import com.google.sps.data.Meal;
```
bprusak: When writing down all the imports, please apply this [rule from Java style guide](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing). Namely:

* Put all static imports into one block, and then all non-static imports into another block,
* Since both exist, put a single empty line between them,
* **There should be no other empty lines between imports.**

Since you need to put everything into a single block, don't forget to sort it alphabetically, by the entire domain (`com.google.appengine.api.datastore.Query` comes before `java.io.IOException`, because `com` comes before `java`).

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,140 @@
...
+    public void tearDown() {
+        helper.tearDown();
+    }
+
+    @Test
+    public void datastoreTest() {
+        DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+        ds.put(new Entity("Meal"));
+        ds.put(new Entity("Meal"));
+        assertEquals(2, ds.prepare(new Query("Meal")).countEntities());
```
bprusak: Remove this test - it was only a proof-of-concept from the tutorial, to show that if the helper is set up properly, Datastore operations can be successfully performed.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.mock.web.MockHttpServletResponse;
+
+public class MealServletTest{
+    private final LocalServiceTestHelper helper =
+        new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+    private final Meal meal_0 = new Meal(
+        1L, "Fried potato", "Fried potato with mushrooms and onion.",
+            new ArrayList<>(Arrays.asList("potato", "onion", "mushrooms", "oil")), "Main");
+    private final Meal meal_1 = new Meal(
```
bprusak: Give these examples some meaningful names, for example - make `MEAL_1` the Meal entry that has its id equal to 1. That's broken for the `meal_0` you have here.
If you want to have two entries with same id to test against failures, name the other something along the lines of `ANOTHER_MEAL_1` or `MEAL_1_DUPLICATE`.
Also, make all of these `private static final` - that will make them constant*. That's also why I suggest to use CONSTANT_CASE instead of lowerCamelCase. (In Java style, there are no lowecase_names_with_underscores. There's either lowerCamelCase, UpperCamelCase or CONSTANT_CASE, or on rare occasions: alllowercase).

* in Java there's no real way to declare a compile-time constant, but `static final` variables are closest to that. `final` means that the reference can't be changed (but any methods can be called - if there are ones modifying content of an object, they will modify that object! Good thing `Meal` doesn't have such methods), `static` means there's only one instance of this object shared between all `MealServletTest` objects ever created.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Test
+    public void getMealByIdTest() throws IOException, ServletException {
+        DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+        ds.put(createMealEntity(meal_2));
+        ds.put(createMealEntity(meal_1));
+        
+        MealServlet servlet = new MealServlet();
+        MockHttpServletRequest request = new MockHttpServletRequest("GET", "/meal/1");
```
bprusak: A `/meal/1` request should get the meal that has id = 1. It's weird to me that the expected result is meal number 2 instead.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -84,13 +77,30 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-        response.setContentType("application/json");
-        response.getWriter().println(gson);
+    private void getMealList(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        Query query = new Query("Meal").addSort("id", SortDirection.DESCENDING);
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery results = datastore.prepare(query);
+        HashMap<Long, Meal> meals = getDataFromDatastore(results);
+        Gson gson = new Gson();
+        response.setContentType("application/json;");
+        response.getWriter().println(gson.toJson(meals));
```
bprusak: You can use `print` instead of `println` if additional end of line in the content is a problem.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,140 @@
...
+import com.google.gson.Gson;
+import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.mock.web.MockHttpServletResponse;
+
+public class MealServletTest{
+    private final LocalServiceTestHelper helper =
+        new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+    private final Meal meal_0 = new Meal(
+        1L, "Fried potato", "Fried potato with mushrooms and onion.",
+            new ArrayList<>(Arrays.asList("potato", "onion", "mushrooms", "oil")), "Main");
```
bprusak: You can make the examples very simplified compared to real world: replace description with "Fried potato description", make every meal have only 1 or 2 ingredients etc.
Unless that's no problem for readability, then let these examples be like they are.

bprusak (changes requested):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,29 +115,67 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+            try {
+                Long id = (Long)entity.getProperty("id");
+                String title = (String) entity.getProperty("title");
+                String description = (String) entity.getProperty("description");
+                ArrayList<String> ingredients = (ArrayList<String>) entity.getProperty("ingredients");
+                String type = (String) entity.getProperty("type");
+                Meal meal = new Meal(id, title, description, ingredients, type);
+                result.put(ind, meal);
+            } catch(ClassCastException e) {
+                continue;
```
zajonc-gugle, line 156: You probably want to at least log the error to `System.out.print`.

In general, whenever an error-ish situation happens, it's better not to silently ignore it. Typically you would either:
1.  propagate the error, either in form of reraising the exception or by including it in the return value (depending on the language)
2.  explicitly handle the error, e.g. by replacing the failing operation with some default
3.  simply log the error message

I'd suggest using 3. here.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,140 @@
...
+        MealServlet servlet = new MealServlet();
+        MockHttpServletRequest request = new MockHttpServletRequest("GET", "/meal");
+        MockHttpServletResponse response = new MockHttpServletResponse();
+
+        try {
+            servlet.doGet(request, response);
+        } catch(Exception e) {
+        }
+        HashMap<Long, Meal> meals = new HashMap();
+        meals.put(0L, meal_1);
```
zajonc-gugle: There's an assumption that keys and values are related, namely `key == value.id`. This relation doesn't hold here, because `meal_1.id == 1`, but it's assigned to key `0`.

Let's try to maintain the assumption in test data and assign `meal_1` to key `1L`, and `meal_2` -> `2L`.

Note, that if you decide to drop the HashMap and use a List in the result (as suggested by Bartek above), the problem will disappear.

zajonc-gugle (changes requested):

pom.xml
polinasand: It requires this dependency, there are errors in tests.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -105,32 +115,71 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+        List<Meal> result = new ArrayList<>();
+        for (Entity entity : query.asIterable()) {
+            try {
+                Long id = (Long)entity.getProperty("id");
+                String title = (String) entity.getProperty("title");
+                String description = (String) entity.getProperty("description");
+                ArrayList<String> ingredients = (ArrayList<String>) entity.getProperty("ingredients");
+                String type = (String) entity.getProperty("type");
+                if (id == null || title.isEmpty() || ingredients.isEmpty()) {
+                    throw new ClassCastException("Invalid value");
```
bprusak: This exception type doesn't match the error you encounter.

There is a multitude of classes dedicated to exceptions (there's hundreds of them), because the class name is part of the message - here `ClassCast` indicates that there was an error while casting from one type into another. It is not really the case when you have issues with null or empty values.
Instead, you should **not** cast an exception in this use case - just rely on `if`/`else` logic to create a meal when data is here, and log a message when it's not.
If you really, *really* want to cast an exception here, create your own exception class, such as `PreconditionFailedException`, `InvalidDatabaseEntryException` etc. It's as simple as creating a new class that just inherits after the `Exception` class - if I remember correctly it doesn't have to have anything new besides constructors.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+        }
+        Gson gson = new Gson();
+        String expected = gson.toJson(MEAL_2);
+        String actual = response.getContentAsString().trim();
+
+        assertEquals(expected, actual);
+    }
+
+    // Get an object from empty datastore.
+    // Result: Response Status NOT FOUND.
```
bprusak: Nit: change the comment to `Expected result: response status "NOT FOUND".`
Add the "Expected" in front of "result" for every test that has this comment.

pom.xml
bprusak: Then please change the scope of the dependency from `core` to `test` if possible.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+        DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+        
+        MealServlet servlet = new MealServlet();
+        MockHttpServletRequest request = new MockHttpServletRequest();
+        MockHttpServletResponse response = new MockHttpServletResponse();
+        request.setPathInfo("/2");
+        
+        try {
+            servlet.doGet(request, response);
+        } catch(Exception e) {
```
bprusak: In this scenario your `doGet()` isn't **expected** to throw exceptions. In such case, you have no need to `try {} catch {}` them. Therefore please remove these blocks and just call `servlet.doGet()` as usual.

If that throws an exception, it will be easier to analyse if you let it crash the test (instead of catching it).

Do this for every test.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+            System.out.println(e);
+        }
+        MockHttpServletResponse expectedResponse = new MockHttpServletResponse();
+        expectedResponse.setStatus(HttpServletResponse.SC_NOT_FOUND);
+        int expected = expectedResponse.getStatus();
```
bprusak: `HttpServletResponse.SC_NOT_FOUND` is just an `int`. So you can discard the `MockHttpServletResponse expectedResponse` entirely, and just do the following:
```
int actual = HttpServletResponse.SC_NOT_FOUND;
```
Or you can even remove this variable too, and just do:
```
assertEquals(HttpServletResponse.SC_NOT_FOUND, response.getStatus());
```

(Please apply this to all tests that expect a status response).

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+        expectedResponse.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+        int expected = expectedResponse.getStatus();
+        int actual = response.getStatus();
+
+        assertEquals(expected, actual);
+    }
+
+    // Get a Meal object from datastore, that can't be created (invalid field value).
+    // Result: catching ClassCastException().
+        @Test
```
bprusak: Fix indentation for this `@Test` statement.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+        MockHttpServletRequest request = new MockHttpServletRequest();
+        MockHttpServletResponse response = new MockHttpServletResponse();
+        request.setPathInfo("/0");
+        
+        try {
+            servlet.doGet(request, response);
+        } catch(ClassCastException e) {
+            System.out.println(e);
+            assertTrue(true);
+        }
```
bprusak: `doGet()` shouldn't throw an exception in this case - both because it's implemented that way *and* because it's good design to carry on even if *some* errors are in the way.

Result I expect: either `SC_NOT_FOUND` or `SC_INTERNAL_SERVER_ERROR`.

Also this test passes whether the exception is caught or not. If exception was thrown, it gets caught here and `assert` passes - it's ok. If the exception is not thrown, then it's not caught, so the `assert` line is never executed. Therefore 0 asserts out of 0 encountered are satisfied, the test passes too.

First and foremost: doGet() shouldn't throw here. Instead it should return a response with an error code.
...
...
But for the future - to actually test if something throws as expected, you need to use `assertThrows()`.

bprusak (changes requested):

pom.xml
polinasand: It needs both test and core, because I already have test.

bprusak (approved):

PR #17, opened 2020-08-28
Restaurants search

grenlayk:
Add search by "type" keyword on map.


src/main/webapp/meal-page-script.js
```diff
@@ -95,7 +95,40 @@ function createMap() {
...
+    };
+    const service = new google.maps.places.PlacesService(map);
+    getSearchPromise(service, request, map).then((results) => {
+        addRestaurants(results, map);
+    })
+  });
+}
+
+function addRestaurants(results, map) {
+    console.log(results);
```
polinasand: This line and 126 line should be removed probably.

polinasand (approved):

zajonc-gugle:
Let's merge master into the branch and resolve conflicts before reviewing the PR.

grenlayk:
Merging done
@zajonc-gugle 
@bprusak

src/main/webapp/meal-page-script.js
```diff
@@ -95,10 +95,39 @@ function createMap() {
...
+      };
+      const service = new google.maps.places.PlacesService(map);
+      getSearchPromise(service, request, map).then((results) => {
+        addRestaurants(results, map);
+      });
     });
 }
 
+function addRestaurants(results, map) {
+  for (let i = 0; i < results.length; i++) {
```
zajonc-gugle: Let's use the `for .. of` syntax here.

src/main/webapp/meal-page-script.js
```diff
@@ -95,10 +95,39 @@ function createMap() {
...
     .then(() => {
-      addRestaurants(map);
+      const request = {
+        location: map.getCenter(),
+        radius: "1500",
+        type: ["restaurant", "cafe", "meal_takeaway", "bar", "bakery"],
+        query: type,
+      };
+      const service = new google.maps.places.PlacesService(map);
+      getSearchPromise(service, request, map).then((results) => {
```
zajonc-gugle: It kind of depends on a use case, but in general it's a good idea to `return` nested promises.

In this case, the top level promise (the one started by `getCurrentPositionPromise` returns nothing. It'll always resolve with `undefined`. If `getSearchPromise` fails, it'll only log to `console.log`, but you won't be able to do anything with that fact.

However, if you `return getSearchPromise(..)` here, the result will be attached to the parent promise. You'd then be able to `catch()` it in line 108, if need be.

I'd suggest:
1. `return`ing the `getSearchPromise` result
2. `catch`ing the result and logging to `console.log` if something went wrong. It's usually a good idea to leave some trace of errors.

zajonc-gugle (approved):

pom.xml
```diff
@@ -73,17 +73,15 @@
...
     </dependency>
     <dependency>
         <groupId>org.springframework</groupId>
-        <artifactId>spring-core</artifactId>
-        <version>4.3.19.RELEASE</version>
-        <scope>core</scope>
+        <artifactId>spring-web</artifactId>
+        <version>5.2.8.RELEASE</version>
     </dependency>
-
```
bprusak: Please undo this change.

This change is good, but it's not part of the frontend work you're doing in this Pull Request. Instead, please put this change in a separate pull request, the one that you're doing on Java backend.

bprusak (approved):

PR #20, opened 2020-08-31
Similar search

grenlayk:


src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -183,6 +183,30 @@ public void getInvalidPathInfoTest() throws IOException, ServletException {
         assertEquals(expected, actual);
     }
 
+    // Get a similar meal from Datastore
+    // Expected result: String in JSON format with id of second meal.
+
```
bprusak: Nitpick: remove the empty line between the comment and the test.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -183,6 +183,30 @@ public void getInvalidPathInfoTest() throws IOException, ServletException {
         assertEquals(expected, actual);
     }
 
+    // Get a similar meal from Datastore
+    // Expected result: String in JSON format with id of second meal.
```
bprusak: Expected result is a number / a Long returned in JSON format, i.e. `2`, not a string in JSON format (`"2"`). Please change this comment accordingly.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -183,6 +183,30 @@ public void getInvalidPathInfoTest() throws IOException, ServletException {
...
+    
+    @Test
+    public void getIdOfSimilarTest() throws IOException, ServletException {
+        DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+        ds.put(createMealEntity(MEAL_1));
```
bprusak, line 190: Add a short comment somewhere (not necessarily here) why meal no. 1 can't be the result (i.e. the "get similar" functionality shouldn't return the same meal as the one the recommendation is based on).

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -36,6 +36,7 @@
 import javax.servlet.ServletException;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
```
bprusak: You are no longer using this annotation, please remove it from imports.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -162,20 +162,24 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
     private void returnIdOfSimilar(HttpServletRequest request, HttpServletResponse response) throws IOException {
         // Fuction redirect to random page
         // TODO(grenlayk): implement suggestions algorithm here for Product Alpha
+        Long pageId = Long.parseLong(getParameter(request, "id", "0"), 10);
```
bprusak: Please skip the second argument (the `10`) - by default `parseLong` parses the string as a decimal number.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -57,7 +57,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         if (pathInfo.split("/", -1).length == 2) {
             String requestType = pathInfo.replaceAll("/", "");
             if (requestType.equals("similar")) {
-                // GET meal/similar
+                // GET meal/similar?id=<pageId>
```
bprusak: Change the comment - it's not a `pageId` you're passing, it's a `meal_id`.
Since you have `meal` objects that you work with, you can refer to them, which makes the comments clearer.

bprusak (changes requested):

polinasand (approved):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -162,20 +162,24 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery results = datastore.prepare(query);
         ArrayList<Long> idList = new ArrayList<>();
         for (Entity entity : results.asIterable()) {
             Long id = (Long)entity.getProperty("id");
-            idList.add(id);
+            if (id != null && !id.equals(mealId)) {
+                idList.add(id);
+            }
         }
```
zajonc-gugle, line 176: Nit:
I believe this block could be rewritten using the [Stream](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html) module.

This looks like a combination on `map` and `filter`.

Not that it matters much, though.

src/main/webapp/meal-page-script.js
```diff
@@ -19,15 +19,15 @@ function fetchMealInfo() {
   const urlParams = new URLSearchParams(queryString);
   const id = urlParams.get("id") ?? 0;
 
-  fetch("/meal/" + id.toString())
+  fetch(`/meal/${id.toString()}`)
```
zajonc-gugle: Nit: you don't need to call `toString()`. It'll be called implicitly in a string template.

`/meal/${id}` would do.

src/main/webapp/meal-page-script.js
```diff
@@ -95,10 +98,39 @@ function createMap() {
...
     .then(() => {
-      addRestaurants(map);
+      const request = {
+        location: map.getCenter(),
+        radius: "1500",
+        type: ["restaurant", "cafe", "meal_takeaway", "bar", "bakery"],
+        query: type,
+      };
+      const service = new google.maps.places.PlacesService(map);
+      getSearchPromise(service, request, map).then((results) => {
```
zajonc-gugle: If you `return` the `getSearchPromise` result here, you'll be able to handle its failures afterwards.

Let's `return getSearchPromise` here and add a `.catch(err => console.log(`Some descriptive message here: ${err}`))` in line 111.

src/main/webapp/meal-page-script.js
```diff
@@ -95,10 +98,39 @@ function createMap() {
...
+        addRestaurants(results, map);
+      });
     });
 }
 
+function addRestaurants(results, map) {
+  for (let i = 0; i < results.length; i++) {
```
zajonc-gugle: Let's use `for .. of` here.

zajonc-gugle (approved):

src/main/webapp/meal-page-script.js
```diff
@@ -107,7 +110,7 @@ function createMap(type) {
       });
     })
     .catch((err) => {
-        console.log(err);
+        console.log(`Catched error: ${err}`);
```
bprusak: Nit: change from "Catched" to "Caught".

bprusak (approved):

PR #21, opened 2020-08-31
Search feature

polinasand:
Implemented a search by words in request form `search-bar`, with spaces between words.
Add some test for this functionality.
Added `<div>` element on `search-results.html` page to show amount of results for concrete search request (@grenlayk review and approve or not).

### DIFF
--- a/src/main/java/com/google/sps/servlets/MealServlet.java
+++ b/src/main/java/com/google/sps/servlets/MealServlet.java
@@ -94,22 +94,29 @@ public class MealServlet extends HttpServlet {
 
     private void getMealList(HttpServletRequest request, HttpServletResponse response) throws IOException {
         List<String> params = Arrays.asList(getParameter(request, "query", "").trim().replaceAll("\\s+", " ").split(" "));
+        params.removeAll(Arrays.asList("\\"));
         Query query = new Query("Meal").addSort("id", SortDirection.ASCENDING);
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery results = datastore.prepare(query);
         List<Meal> meals = getDataFromDatastore(results);
-        List<Meal> searchedMeal = new ArrayList<>();
+
+        // In case if search request is empty.
+        if (params.isEmpty()) {
+            response.setContentType("application/json;");
+            response.getWriter().print(new Gson().toJson(meals));
+            return;
+        }
 
         // Every object Meal adds if it contains at least one keyword in at least one field 
         // (type, title, description, ingredients).
+        List<Meal> searchedMeal = new ArrayList<>();
         for (Meal meal : meals) {
-            if (!searchedMeal.contains(meal) && isResultOfSearch(meal, params)) {
+            if (isResultOfSearch(meal, params)) {
                 searchedMeal.add(meal);
             }
         }
-        Gson gson = new Gson();
         response.setContentType("application/json;");
-        response.getWriter().print(gson.toJson(searchedMeal));
+        response.getWriter().print(new Gson().toJson(searchedMeal));
         return;
     }
 
--- a/src/main/java/com/google/sps/servlets/MealServlet.java
+++ b/src/main/java/com/google/sps/servlets/MealServlet.java
@@ -29,7 +29,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Enumeration;
 import java.util.List;
 import java.util.Random;
 import java.util.stream.Collectors;
@@ -49,6 +48,7 @@ public class MealServlet extends HttpServlet {
 
         // Get meal
         // GET meal/
+        // GET meal?query=<smth>
         if (pathInfo == null || pathInfo.equals("/")) {
             getMealList(request, response);
             return;
@@ -59,9 +59,6 @@ public class MealServlet extends HttpServlet {
             if (requestType.equals("similar")) {
                 // GET meal/similar
                 returnIdOfSimilar(request, response);
-            } else if (requestType.contains("query=")) { 
-                // GET meal/query=
-                getMealList(request, response);
             } else {
                 // GET meal/<meal_id>
                 getMealById(request, response);
@@ -96,9 +93,7 @@ public class MealServlet extends HttpServlet {
     }
 
     private void getMealList(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        // Search by full query request.
-        String p = getParameter(request, "query", "");
-        List<String> params = new ArrayList<>(Arrays.asList(p));
+        List<String> params = Arrays.asList(getParameter(request, "query", "").trim().replaceAll("\\s+", " ").split(" "));
         Query query = new Query("Meal").addSort("id", SortDirection.ASCENDING);
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery results = datastore.prepare(query);
@@ -108,19 +103,13 @@ public class MealServlet extends HttpServlet {
         // Every object Meal adds if it contains at least one keyword in at least one field 
         // (type, title, description, ingredients).
         for (Meal meal : meals) {
-            for (String param : params) {
-                if (!searchedMeal.contains(meal) && 
-                    (meal.getTitle().contains(param) ||
-                    meal.getDescription().contains(param) ||
-                    meal.getType().contains(param) ||
-                    meal.getIngredients().contains(param))) {
-                        searchedMeal.add(meal);
-                    }
+            if (!searchedMeal.contains(meal) && isResultOfSearch(meal, params)) {
+                searchedMeal.add(meal);
             }
         }
         Gson gson = new Gson();
         response.setContentType("application/json;");
-        response.getWriter().println(gson.toJson(searchedMeal));
+        response.getWriter().print(gson.toJson(searchedMeal));
         return;
     }
 
@@ -196,10 +185,29 @@ public class MealServlet extends HttpServlet {
         Long randomId = idList.get(index);
         Gson gson = new Gson();
         response.setContentType("application/json;");
-        response.getWriter().println(gson.toJson(randomId));
+        response.getWriter().print(gson.toJson(randomId));
         return;
     }
 
+    private Boolean isResultOfSearch(Meal meal, List<String> params) {
+        for (String param : params) {
+            if (meal.getTitle().contains(param)) {
+                return true;
+            } 
+            if (meal.getDescription().contains(param)) {
+                return true;
+            } 
+            if (meal.getType().contains(param)) {
+                return true;
+            }
+            for (String ingredient : meal.getIngredients()) {
+                if (ingredient.contains(param)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
     private String getParameter(HttpServletRequest request, String name, String defaultValue) {
         String value = request.getParameter(name);
         if (value == null) {
--- a/src/main/java/com/google/sps/servlets/MealServlet.java
+++ b/src/main/java/com/google/sps/servlets/MealServlet.java
@@ -96,12 +96,9 @@ public class MealServlet extends HttpServlet {
     }
 
     private void getMealList(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        String pathInfo = request.getPathInfo();
-        if (pathInfo == null) {
-            pathInfo = "";
-        }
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -29,6 +29,7 @@
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Enumeration;
```
grenlayk: Better put `import java.util.Enumeration;` one row above

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -59,7 +59,10 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
             if (requestType.equals("similar")) {
                 // GET meal/similar
                 returnIdOfSimilar(request, response);
-            } else { 
+            } else if (requestType.contains("query=")) { 
+                // GET meal/query=
```
grenlayk: I suppose `query` should be parameter:
```
// GET meal?query=<something>
```

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -93,13 +96,34 @@ private Entity createMealEntity(Meal meal) {
...
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery results = datastore.prepare(query);
         List<Meal> meals = getDataFromDatastore(results);
+        List<Meal> searchedMeal = new ArrayList<>();
+
+        // Every object Meal adds if it contains at least one keyword in at least one field 
+        // (type, title, description, ingredients).
+        for (Meal meal : meals) {
+            for (String param : params) {
+                if (!searchedMeal.contains(meal) &&
```
grenlayk: Probably it will be better to create func `isRelated` for example, in which you will have all this conditions.

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: I don't see why would we need to move `Enumeration`, right now it's in alphabetical order (Collections, Enumeration, List).

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: Agreed. If the request is `/meal?query=<something>`, then the pathInfo will be set to `"/meal"`, and you will be able to `request.getParameter("query")`, which should return `"<something>"`.

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: It's a good idea to put this big condition into a function that will act as a predicate. It should run against arguments `Meal` and `Collection<String>`, so that the check against multiple query params is contained within the function (instead of using `.contains` to check if the meal already classified as matching the query. `contains` for an `List` is of linear complexity by the way, so we should avoid it to not increase the complexity of the search algorithm).

In a perfect solution, we'd use Elasticsearch to actually perform the search, instead of our own application filtering the results out. However, this requires learning new API, some tricky setup (besides starting an Elasticsearch engine some workarounds for the API is needed on App Engine) and a lot of extra effort that we might not afford (especially if we want some other features to be implemented before internship ends).

So let's keep it this way, but keep in mind - in a perfect world, such filtering would be performed by the database (or some dedicated search engine) to make the search efficient (i.e. not browse the entire database every time, but use indexes to optimize the search by a *lot*).

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -93,13 +96,34 @@ private Entity createMealEntity(Meal meal) {
...
+                    meal.getType().contains(param) ||
+                    meal.getIngredients().contains(param))) {
+                        searchedMeal.add(meal);
+                    }
+            }
+        }
         Gson gson = new Gson();
         response.setContentType("application/json;");
-        response.getWriter().print(gson.toJson(meals));
+        response.getWriter().println(gson.toJson(searchedMeal));
```
bprusak: Use `print` instead of `println` for output Json.

src/main/webapp/search-meal-script.js
```diff
@@ -13,9 +13,14 @@
 // limitations under the License.
 
 function searchMeal() {
-  fetch("/meal")
+  const queryString = window.location.search;
+  const urlParams = new URLSearchParams(queryString);
+  const searchRequest = urlParams.get("query") ?? "";
+  fetch(`/meal/query=${searchRequest}`)
```
bprusak: Reminder: please use `/meal?query=<request>` instead.

src/main/webapp/main.js
```diff
@@ -12,15 +12,15 @@ Vue.component("search-bar", {
...
+      <div v-on:click="submit" type=submit id="search-icon"></div>
     </div>
   </div>
   `,
   methods: {
     submit: function () {
-      window.location.replace("/search-results.html");
-      searchMeal();
+      const searchLine = document.getElementById('query').value;
+      window.location.replace('search-results.html?query=' + searchLine);
```
bprusak: It's possible that `searchLine` needs to have some special URI encoding (like replacing spaces and special characters in the "query" string with "%2F", "+" or other special sequences). This might happen automatically, but if not - use `encodeURIComponent()` to do so (as in `` `search-results.html?query=${encodeURIComponent(searchLine)}` ``)

src/main/webapp/search-meal-script.js
```diff
@@ -13,9 +13,14 @@
...
     .then((response) => response.json())
     .then((dishes) => {
+      const amount = document.getElementById('amount-block');
+      amount.innerText = `${dishes.length} results`;
```
bprusak: The word "results" should become "result" when there's exactly 1 result.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -154,6 +154,7 @@ public void getMealListTest() throws IOException, ServletException {
         MealServlet servlet = new MealServlet();
         MockHttpServletRequest request = new MockHttpServletRequest();
         MockHttpServletResponse response = new MockHttpServletResponse();
+        request.setPathInfo("/query=");
```
bprusak: In the test, to simulate `/meal?query=aaa`, do `setPathInfo("/")` and `addParameter("query","aaa")`.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -165,6 +166,50 @@ public void getMealListTest() throws IOException, ServletException {
...
+        DatastoreService ds = DatastoreServiceFactory.getDatastoreService();
+        ds.put(createMealEntity(MEAL_1));
+        ds.put(createMealEntity(MEAL_2));
+        
+        MealServlet servlet = new MealServlet();
+        MockHttpServletRequest request = new MockHttpServletRequest();
+        MockHttpServletResponse response = new MockHttpServletResponse();
+        request.setPathInfo("/query=potato onion");
+        servlet.doGet(request, response);
+        List<Meal> meals = new ArrayList<>();
```
bprusak, line 182: Put the empty line before `List<Meal>` and remove one before `Gson gson = new Gson();`.

That way the empty lines (together with the one earlier) split the test into 3 typical steps of a test: Arrange, Act, Assert. (This `List` is part of preparing the assertion, and is not related to the Act part that is the `doGet()` call.)

Please do that change to other tests too.

bprusak (changes requested):

src/main/java/com/google/sps/servlets/MealServlet.java
polinasand: Requests dont work when I use passing parameters, multiple calls of `doGet()` stopped only after I changed endpoint. 
For endpoint /meal function `searchMeal()` in the `body onload` fetches wrong data.

src/main/java/com/google/sps/servlets/MealServlet.java
grenlayk: Yep, sorry

src/main/webapp/search-meal-script.js
zajonc-gugle: The simple way to do it is to just use the ternary operator:
```
const count = dishes.length;
const isSingular = count == 1;
const text = `${dishes.length} result${isSingular? "" : "s"}`;
```

The cleaner approach would be to extract it to a function.

The most future-proof solution would be to use an external library: https://github.com/plurals/pluralize
You can load the script with a regular `script src` tag, using this url as the path: https://raw.githubusercontent.com/plurals/pluralize/master/pluralize.js

However, I think either of the simpler approaches would do. There's no point introducing a library just to append an `"s"` once.

zajonc-gugle (changes requested):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -93,13 +93,23 @@ private Entity createMealEntity(Meal meal) {
...
         Query query = new Query("Meal").addSort("id", SortDirection.ASCENDING);
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery results = datastore.prepare(query);
         List<Meal> meals = getDataFromDatastore(results);
+        List<Meal> searchedMeal = new ArrayList<>();
+
+        // Every object Meal adds if it contains at least one keyword in at least one field 
+        // (type, title, description, ingredients).
+        for (Meal meal : meals) {
+            if (!searchedMeal.contains(meal) && isResultOfSearch(meal, params)) {
```
bprusak: Do NOT perform this `.contains()` call - each meal appears only once on the `meals` list and there's no way you'll add the same meal twice to the results.

The `contains` hurts performance in this scenario, and is completely unnecessary.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -154,11 +153,56 @@ public void getMealListTest() throws IOException, ServletException {
         MealServlet servlet = new MealServlet();
         MockHttpServletRequest request = new MockHttpServletRequest();
         MockHttpServletResponse response = new MockHttpServletResponse();
+        request.setParameter("query", "");
```
bprusak: Remove `setParameter("query",...)` from this particular test

When someone calls `/meal` just to get a list of all the meals, then they won't set the "query" parameter to anything. Your code should be able to handle this situation, if it doesn't - the app code needs fixing, not the test.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -93,13 +93,23 @@ private Entity createMealEntity(Meal meal) {
     }
 
     private void getMealList(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        List<String> params = Arrays.asList(getParameter(request, "query", "").trim().replaceAll("\\s+", " ").split(" "));
```
bprusak, line 96: From this list remove any entry that is an empty string.

Without this, any query that has multiple spaces inside (such as "potato&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;onion") will appear in `params` list as a lot of strings, many of them being empty (i.e. zero-charater `""` string), i.e. it will be 
```
{"potato",
"",
"",
...
"",
"",
"onion"}
```
The empty strings will match any string, and therefore if one searches "potato&nbsp;&nbsp;&nbsp;onion" they will get all the entries, instead of only ones with "potato" or "onion".

Once you fix this and remove all empty strings from `params` list, you'll find that the `/meal` request no longer returns all meals - that's because it relied on `params` having the empty string entry. Instead, you should add explicit logic that when the "query" is absent from the request, or when `params` list is empty, then no filtering should be done and all meals should be returned instead.

src/main/webapp/main.js
```diff
@@ -11,16 +11,16 @@ Vue.component("photo-copyright", {
 Vue.component("search-bar", {
   template: `
   <div id="search-bar">
-    <div id="bar">
-      <input v-on:keyup.enter="submit" v-model="request" type="text" id="text" name="text" placeholder="Search..." />
-      <div v-on:click="submit" id="search-icon"></div>
-    </div>
+    <form id="bar" action="/meal">
```
bprusak: Please undo the `form`, the `action` and the `type=submit`.

This action is done via Javascript already by the means of `fetch` function. Adding it here adds a chance some implementation goes wrong and the request to `/meal` gets submitted twice: once here, and in `fetch` in javascript later.

bprusak (changes requested):

zajonc-gugle (approved):

PR #22, opened 2020-09-03
Change dependencies

grenlayk:
My project did not compile with previous dependencies, so probably we can update them.
@sandatsian please check whether it works for you or not (do not forget about `mvn clean` before compilation)


bprusak (approved):

polinasand (approved):
My project compiles with this, too.

PR #26, opened 2020-09-08
Wiki Import

polinasand:
Implemented ability to add Meal entity to Datastore from web-page (now [https://en.wikibooks.org/wiki/Cookbook:Racuszki]). 
Parse text from web page in `script.js` and send request to `WikiImportServlet` as JSON String, where put new `Meal` object  to Datastore.


src/main/webapp/script.js
```diff
@@ -18,3 +18,48 @@ function createElementByTag(text, tag) {
...
+/**
+ * parse.js
+ *
+ * MediaWiki API Demos
+ * Demo of `Parse` module: Parse content of a page
+ *
+ * MIT License
+ */
+
+async function parseData() {
```
grenlayk: Everywhere we use promises. Maybe it is better to use them here too instead of `async` & `await`?

src/main/webapp/script.js
```diff
@@ -18,3 +18,48 @@ function createElementByTag(text, tag) {
   element.innerText = text;
   return element;
 }
+
+/**
+ * parse.js
```
grenlayk: `script.js` was created to write functions which are used for several `.js` files.
Maybe it will be better to add all this data in `parse.js`?

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,73 @@
...
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        datastore.put(createMealEntity(meal));
+    }
+
+    /**
+     * Creates an entity for Datastore with properties of class Meal.
+     * @param meal object of class Meal for which the entity is creating.
+     * @return new Entity object with necessary properties.
+     */
+    private Entity createMealEntity(Meal meal) {
```
grenlayk: We already have such function. @bprusak maybe there is a way to reuse it from `MealServlet.java`?

bprusak: To do that, it's best to move this method into another class (that inherits nothing, as opposed to being yet another servlet). It needs to be set to `public`, and preferably `static`.
If the method is static and put into a class named, let's say `MyDataConversionTools`, you can import that class and then call `MyDataConversionTools.createMealEntity()`.
For such utilities it's best to do static methods, because otherwise you'd need to create an object first (`MyDataConversionTools conversion = new MyDataConversionTools(); conversion.createMealEntity();`). The latter is useful if some more context is needed than just the method arguments.

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.gson.Gson;
+import com.google.sps.data.Meal;
+
+@WebServlet("/import")
+public class WikiImportServlet extends HttpServlet {
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException{
```
bprusak: I believe it's better to send and receive via POST, then putting and retrieving the content should be simple - you put them as "body", and read it here using `request.getReader()` that lets you read in any way you like - line by line, entire string in one go etc.
Sending data (especially to upload them to database) via POST is the mainstream way of doing it.

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,73 @@
...
+        String json = request.getQueryString().replaceAll("input=", "");
+        try {
+            json = URLDecoder.decode(json);
+        } catch (Exception e) {  
+            System.out.println("Issue while decoding" +e.getMessage());
+            return;  
+        } 
+        // Checking if JSON String is correct.
+        System.out.println(json);
```
bprusak: This println() isn't needed in the final product, please remove it.

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
polinasand: Okay I'll create such a class.

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,73 @@
...
+        Meal meal = new Gson().fromJson(json, Meal.class);
+        System.out.println(meal.getTitle());
```
bprusak: This message might be useful, but:
1) you need to add some description so that the log can be undestood without knowing this particular line of code, such as "Adding meal titled: \"" + meal.getTitle() + "\""
2) use a logger instead of `System.out.println`. The [App Engine tutorial](https://cloud.google.com/appengine/docs/standard/java/logs) says to just use `java.util.logging.Logger`. The usual way to use it is to add a static field in your class, that should be defined like this: `private static final Logger log = Logger.getLogger(WikiImportServlet.class.getName());`. There should be one such logger in every class. In case of this message, you'd want to use `info` method: `logger.info("Adding meal titled: " + meal.getTitle())`

src/main/webapp/script.js
```diff
@@ -18,3 +18,48 @@ function createElementByTag(text, tag) {
...
+    const data = JSON.stringify(
+        {
+            id: 25,
+            title: "racuszki",
+            description: "Racuszki are pancakes made from batter fried on a hot griddle or a pan. They are a traditional Polish meal.",
+            type: "Pancake",
+            ingredients: ingredients
+        });
+     
+    window.location.replace("/import?input="+encodeURIComponent(data));
```
bprusak: Please use `fetch` instead of `window.location.replace`. This will make things easier, especially if you'll want to process multiple Wikipedia articles at a time.
Also, `fetch` has a convenient `body:` parameter where you can put the JSON you want to send.

bprusak (changes requested):

src/main/webapp/script.js
```diff
@@ -18,3 +18,48 @@ function createElementByTag(text, tag) {
...
+    const json = await request.json();
+    const container = document.getElementById("data-container");
+    container.innerHTML = json.parse.text['*'];
+
+    const ingredients = [];
+    const ingrHTML = document.getElementsByTagName("ul").item(0).children;
+    console.log(ingrHTML);
+    Array.from(ingrHTML).forEach((el) => {
+        ingredients.push(el.innerText);
+        });
```
zajonc-gugle: A slightly more idiomatic way to write this would be:
```
const ingredients = ingrHTML.map(el => el.innerText);
```

src/main/webapp/script.js
zajonc-gugle: Just a hint: async / await is just syntactic sugar for promises.

Removing the `await / async` keywords should be quite simple. It'll only require using `then` where you use `await` now, and adding a few extra curly braces.

src/main/webapp/import.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<head>
+    <meta charset="UTF-8" />
+    <title>Dataset</title>
+    <script src="script.js"></script>
+</head>
+<body>
+    <div id="data-container"></div>
+        <button type=submit onclick="parseData()">Import</button>
```
zajonc-gugle: Idk if this is intended, but `button` is outside of `div`.

src/main/webapp/script.js
```diff
@@ -18,3 +18,48 @@ function createElementByTag(text, tag) {
...
+    new URLSearchParams({
+        origin: "*",
+        action: "parse",
+        page: "Cookbook:Racuszki",
+        format: "json",
+        prop: "text",
+    });
+
+    const request = await fetch(url);
+    const json = await request.json();
```
zajonc-gugle: I'd suggest not using `json` as a variable name. `JSON` is a global object in JS. Calling a variable `json` is a bit confusing.

Perhaps `body`?

zajonc-gugle (changes requested):

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,60 @@
...
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String json = IOUtils.toString(request.getReader());
+        try {
+            json = URLDecoder.decode(json);
+        } catch (Exception e) {  
+            System.out.println("Issue while decoding" +e.getMessage());
+            return;  
+        } 
+        Meal meal = new Gson().fromJson(json, Meal.class);
+        System.out.println("meal " + meal.getTitle());
```
bprusak: Use a `java.util.logging.Logger` to print these messages, instead of `System.out`.

You need to create a class member `private static final Logger log = Logger.getLogger(WikiImportServlet.class.getName());`
Then here in code, you need to use `log.info("........")`

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/import")
+public class WikiImportServlet extends HttpServlet {
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String json = IOUtils.toString(request.getReader());
+        try {
+            json = URLDecoder.decode(json);
```
bprusak: Please remove this conversion and just use the string you got from `request.getReader()`.

If the string came in as body, then it was never URL-encoded, so you don't need to do any decoding here.

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/import")
+public class WikiImportServlet extends HttpServlet {
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
bprusak, line 50: Make a unit test for this functionality. I think of 2 scenarios: simple successful one, and one where JSON is malformed (such as `title:"xxx"}` - some kind of syntax error).

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,60 @@
...
+        String json = IOUtils.toString(request.getReader());
+        try {
+            json = URLDecoder.decode(json);
+        } catch (Exception e) {  
+            System.out.println("Issue while decoding" +e.getMessage());
+            return;  
+        } 
+        Meal meal = new Gson().fromJson(json, Meal.class);
```
bprusak: Please use `try { Gson().fromJson(...); } catch(JsonSyntaxException e) { ... }` here, in expectation that JSON message received may have some errors.

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,60 @@
...
+
+@WebServlet("/import")
+public class WikiImportServlet extends HttpServlet {
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String json = IOUtils.toString(request.getReader());
+        try {
+            json = URLDecoder.decode(json);
+        } catch (Exception e) {
```
bprusak: Using very general exceptions for `catch` is a bad habit, please try to be more specific when possible.
Otherwise you might catch an exception that's a fatal error in your app that urgently needs debugging, and instead it gets stifled here.

src/main/webapp/parse.js
```diff
@@ -0,0 +1,70 @@
...
+function sendData() {
+    var url;
+    try {
+        url = "https://en.wikibooks.org/w/api.php?" +
+        new URLSearchParams({
+            origin: "*",
+            action: "parse",
+            page: `Cookbook:${document.getElementById("input-text").value}`,
+            format: "json",
+            prop: "text",
```
bprusak, line 32: Consider the following for a future pull request:
"text" returns the parsed HTML page - you take advantage of it by loading the HTML and navigating it using `document.getElementsBy*`.
However, did you consider using "wikitext" instead? It may not have an HTML format, but it's simpler to analyze, to the point that you can process the article by searching the text with `.match()` function (for such keywords as "==Ingredients==").
An example of a wikitext response is [here](https://en.wikibooks.org/w/api.php?action=parse&page=Cookbook:Racuszki&format=json&prop=wikitext).

bprusak (changes requested):

src/main/webapp/parse.js
polinasand: The wikitext format requires additional processing like remove extra symbols, for example if "batter" is a link and it represented like 
[[Cookbook:Batter|batter]]
Thats why I use text, however some HTML tags on wiki pages don't have ElementId, and I should check if webpage has correct structure.

grenlayk (approved):

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,82 @@
...
+@WebServlet("/import")
+public class WikiImportServlet extends HttpServlet {
+
+    private static final Logger logger = Logger.getLogger(WikiImportServlet.class.getName());
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String json = IOUtils.toString(request.getReader());
+        try {
+            System.out.println(json);
```
bprusak: Either change it to `logger.info` or remove it, you shouldn't have `System.out.print` in your final code for the Servlet.

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
```diff
@@ -0,0 +1,82 @@
...
+                new Meal(
+                    id,
+                    meal.getTitle(),
+                    meal.getDescription(),
+                    meal.getIngredients(),
+                    meal.getType()
+                )
+            ));
+        } catch (JsonSyntaxException e) {
+            System.out.println(e);
```
bprusak: Either change it to `logger.warning` or remove it, you shouldn't have `System.out.print` in your final code for the Servlet.

src/test/java/com/google/sps/WikiImportTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @After
+    public void tearDown() {
+        helper.tearDown();
+    }
+
+    // Put one object Meal to empty Datastore, created from valid JSON String.
+    // Expected result: Datastore has one entity for this object Meal with id = 0L.
+    @Test
+    public void putValidObjectTest() throws IOException, ServletException {
```
bprusak: Add one more test: if there's an existing entry with some ID in the Datastore, the newly submitted entry should have an ID that's 1 higher.

bprusak (approved):

src/main/java/com/google/sps/wikiimport/WikiImportServlet.java
polinasand: As always I forgot to remove lines like this.

PR #28, opened 2020-09-09
Render form value with Vue

zajonc-gugle:
I'm not sure if this solves the problem, but the changes seem to keep the form value in sync with the query using Vue.

Note, that it's not an ideal approach (as described in a comment), but I'd say it's a step towards the "proper" approach, i.e. calculating things in Vue and passing values to the template.

I'm still new to Vue, so there may be some non-idiomatic pieces there, but it seems to do the trick.

All of that assuming that the problem was to find a way to replace `document.getElementById('query').setAttribute` with a more Vue-ish approach.

### DIFF
--- a/src/main/webapp/main.js
+++ b/src/main/webapp/main.js
@@ -18,7 +18,7 @@ Vue.component("search-bar", {
   template: `
   <div id="search-bar">
     <div id="bar">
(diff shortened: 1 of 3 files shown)


bprusak (approved):
I tested it briefly and it seems to work well!

grenlayk (approved):
Well, this works for me too
Thanks!

PR #29, opened 2020-09-10
Update getSimilar func

grenlayk:


src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -177,23 +177,45 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
         Query query = new Query("Meal");
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         PreparedQuery results = datastore.prepare(query);
+
+        for (Entity entity : results.asIterable()) {
+            Long id = (Long)entity.getProperty("id");
+            if (id != null && id.equals(mealId)) {
+                mealType = (String)entity.getProperty("type");
+            }
+        }
```
bprusak: Instead of requesting all data and going through it one by one, please `addFilter` to the Query that will filter out by id being equal to `mealId`. You can look up `getMealById` for how to do that.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -177,23 +177,45 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
         ArrayList<Long> idList = new ArrayList<>();
         for (Entity entity : results.asIterable()) {
             Long id = (Long)entity.getProperty("id");
+            String type = (String)entity.getProperty("type");
             if (id != null && !id.equals(mealId)) {
                 idList.add(id);
             }
+            if (id != null && !id.equals(mealId) && type != null && type.equals(mealType)) {
+                sameTypeIdList.add(id);
+            }
```
bprusak: This filtering can also be done via query filters:
Make a separate query that has a filter on "type" to be equal to `mealType` and a filter on "id" to be NOT equal to `mealId`.
Offloading all these operations onto the database instead of the application is preffered, because database can do this efficiently and there's less data to be streamed and held by the servlet itself.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -177,23 +177,45 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
         }
+
+
         Random rand = new Random(); 
         int index = rand.nextInt(idList.size());
         Long randomId = idList.get(index);
```
bprusak: Request the entire list of entries from the database to take a random one only *after* it turns out there's no truly similar meals found.

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -249,6 +256,31 @@ public void getIdOfSimilarTest() throws IOException, ServletException {
...
+        MealServlet servlet = new MealServlet();
+        MockHttpServletRequest request = new MockHttpServletRequest();
+        MockHttpServletResponse response = new MockHttpServletResponse();
+        request.setPathInfo("/similar");
+        request.addParameter("id", PIZZA_1.getId().toString());
+        servlet.doGet(request, response);
+
+        // returnIdOfSimilar() should  return id of the same category meel
+        Long similarId = PIZZA_2.getId();
+
```
bprusak: Remove this empty line.

Too many empty lines makes the code harder to read, as it is artificially spread out. Empty lines should be used to split code like splitting text into paragraphs. In tests we usually split into setup, action and expectations (and this split is achieved in this test by the other 2 empty lines).

bprusak (changes requested):

src/test/java/com/google/sps/MealServletTest.java
```diff
@@ -228,7 +235,7 @@ public void getInvalidPathInfoTest() throws IOException, ServletException {
     // Get a similar meal from Datastore
     // Expected result: Long in JSON format with id of second meal.
     @Test
-    public void getIdOfSimilarTest() throws IOException, ServletException {
+    public void noSameTypeTest() throws IOException, ServletException {
```
polinasand, line 239: You can add one more test when Datastore contains the only one entity.

polinasand (approved):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -177,26 +179,47 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+        List<Meal> idList = getDataFromDatastore(datastore.prepare(idQuery));
+
         Random rand = new Random(); 
-        int index = rand.nextInt(idList.size());
-        Long randomId = idList.get(index);
+        int index = 0;
+        Meal randomMeal = null;
+        if (sameTypeIdList.size() > 0) {
+            index = rand.nextInt(sameTypeIdList.size());
+            randomMeal = (Meal)sameTypeIdList.get(index);
```
zajonc-gugle: I guess you can write a function that would just pick a random item from a list.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -177,26 +179,47 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+        } else if (idList.size() > 0) {
+            index = rand.nextInt(idList.size());
+            randomMeal = (Meal)idList.get(index);
```
zajonc-gugle: We only use `idList` in this `else` clause, but we fetch the results it in all cases.
We could optimize the logic a bit by delaying evaluation of the `idList` until we actually need it. If you move the `idList = ...` assignment to an `else` block here, you'll avoid an unnecessary request to the database if `sameTypeIdList.size() > 0`.

zajonc-gugle (approved):

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -177,26 +179,47 @@ private void getMealById(HttpServletRequest request, HttpServletResponse respons
...
+        Query idQuery = new Query("Meal").setFilter(anotherIdFilter);
+        List<Meal> sameTypeIdList = getDataFromDatastore(datastore.prepare(typeQuery));
+        List<Meal> idList = getDataFromDatastore(datastore.prepare(idQuery));
+
         Random rand = new Random(); 
-        int index = rand.nextInt(idList.size());
-        Long randomId = idList.get(index);
+        int index = 0;
+        Meal randomMeal = null;
+        if (sameTypeIdList.size() > 0) {
```
bprusak: Please use `.isEmpty()` instead of checking if the size is 0.

In general checking emptiness with the dedicated method is much faster than getting (or calculating) the size of a collection.

src/main/java/com/google/sps/servlets/MealServlet.java
bprusak: +1 for this idea. By "evaluation of the `idList`" Tomek means to move this entire line over here: `List<Meal> idList = getDataFromDatastore(datastore.prepare(idQuery));` 

In a perfect world your app would never need to retrieve the entire database of entries to work on it. Getting all the entries is expensive both in terms of transferring all the data between Datastore and your app, as well as it may take a lot of RAM to keep all the downloaded data.

bprusak (changes requested):

bprusak (approved):

PR #31, opened 2020-09-10
Some tiny things about Wiki Import

polinasand:
This PR for `type` initialization (by user in import.html) and some issues with decoding strings.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,14 @@
         <version>5.2.8.RELEASE</version>
         <scope>core</scope>
     </dependency>
-      
+
+    <!-- https://mvnrepository.com/artifact/commons-io/commons-io -->
+    <dependency>
+        <groupId>commons-io</groupId>
+        <artifactId>commons-io</artifactId>
+        <version>2.8.0</version>
+    </dependency>
+
   </dependencies>
 
   <build>
--- a/src/main/webapp/import.html
+++ b/src/main/webapp/import.html
@@ -7,6 +7,33 @@
 <body>
     <div>
         <input name="input" type="text" id="input-text"/>
+        <select id="type">
+            <option>beverage</option>
+            <option>bread</option>
+            <option>casserole</option>
+            <option>chili</option>
+            <option>cocktails</option>
+            <option>dessert</option>
+            <option>fish</option>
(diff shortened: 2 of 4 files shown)


src/main/webapp/import.html
```diff
@@ -7,6 +7,33 @@
 <body>
     <div>
         <input name="input" type="text" id="input-text"/>
+        <select id="type">
```
bprusak, line 10: Is this particular list based on some recipes you've already gathered and want to import? Or did you just make it from memory?

pom.xml
```diff
@@ -83,7 +83,14 @@
         <version>5.2.8.RELEASE</version>
         <scope>core</scope>
     </dependency>
-      
+
+    <!-- https://mvnrepository.com/artifact/commons-io/commons-io -->
+    <dependency>
+        <groupId>commons-io</groupId>
```
bprusak, line 89: Please don't add the library if there's no code using it.
Maybe you forgot to add Java code to this pull request?

bprusak (changes requested):

pom.xml
polinasand: I forgot to add this change to previous PR.
I should add it separately from this changes, I think.

src/main/webapp/import.html
polinasand: This list based on https://en.wikibooks.org/wiki/Category:Recipes_by_type list of types, excluding some repeated types or not so popular.

src/main/webapp/search-meal-script.js
```diff
@@ -46,7 +46,7 @@ function createMealElement(dish) {
   const aElement = document.createElement("a");
   aElement.setAttribute("href", `meal.html?id=${id}`);
   const insideDivElement = document.createElement("div");
-  insideDivElement.appendChild(createElementByTag(title, "b"));
+  insideDivElement.appendChild(createElementByTag(capitalizeFirst(title), "b"));
```
grenlayk, line 49: I guess I found where problem with encoding appears.
I guess we will need to add `encodingCheck()` here too as I did it in `meal-page-script.js`. 
I made PR to add this func to `script.js` file.

PR #34, opened 2020-09-14
Issues found while filling Datastore

polinasand:
In `parse.js`: add case when there is no "description" on the page.
In `script.js`: Change capitalize function to capitalize first letter of each word in title.
In `meal-page-script.js`: remove capitalizing first letters of ingredients (according to Wikipedia and others sites format of presenting the list of ingredients).

### DIFF
--- a/src/main/webapp/parse.js
+++ b/src/main/webapp/parse.js
@@ -23,7 +23,7 @@
 function sendData() {
     var url;
     // Get type of meal.
-    var typeSelector = document.getElementById("type");
+    const typeSelector = document.getElementById("type");
     var type = typeSelector.options[typeSelector.selectedIndex].innerText ?? "";
     try {
         url = "https://en.wikibooks.org/w/api.php?" +
@@ -44,14 +44,11 @@ function sendData() {
             const container = document.getElementById("data-container");
             container.innerHTML = jsonData.parse.text['*'];
             try {
-                const ingrHTML = document.getElementsByTagName("ul").item(0).children;
+                const ingrHtml = document.getElementsByTagName("ul").item(0).children;
                 const list = document.getElementsByClassName("mw-parser-output")[0];
-                const ingredients = Array.from(ingrHTML).map(el => el.innerText);
-                const descriptionHTML = list.getElementsByTagName("p")[1];
-                var description = "";
-                if (descriptionHTML != null) {
-                    description = descriptionHTML.textContent ?? "";
-                }
+                const ingredients = Array.from(ingrHtml).map(el => el.innerText);
+                const descriptionHtml = list.getElementsByTagName("p")[1];
+                const description = descriptionHTML == null ? "" : descriptionHtml.textContent ?? "";
                 const title = jsonData.parse.title.replace("Cookbook:", "").toLowerCase();
                 // Create JSON String for Meal object.
                 // By default id of object is 0, before putting object to Datastore
--- a/src/main/webapp/parse.js
+++ b/src/main/webapp/parse.js
@@ -22,7 +22,9 @@
 
 function sendData() {
     var url;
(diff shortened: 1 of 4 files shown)


src/main/webapp/script.js
```diff
@@ -20,7 +20,12 @@ function createElementByTag(text, tag) {
 }
 
 function capitalizeFirst(string) {
-  return string.charAt(0).toUpperCase() + string.slice(1).toLowerCase();
+  var result = string
```
grenlayk: Let's use `const` instead of `var`

src/main/webapp/script.js
```diff
@@ -20,7 +20,12 @@ function createElementByTag(text, tag) {
...
+    .split(' ')
+    .map(word => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(' ');
+  console.log(result);
```
grenlayk: Do we still need log?

src/main/webapp/parse.js
```diff
@@ -22,7 +22,9 @@
 
 function sendData() {
     var url;
-    const type = document.getElementById("type").innerText ?? "";
+    // Get type of meal.
+    var typeSelector = document.getElementById("type");
```
grenlayk: Let's not use `var`

bprusak (approved):

src/main/webapp/script.js
zajonc-gugle: Or, since we return the result right away (assuming console.log is to be removed anyway), let's just `return` right away, without assigning the value to a temporary variable `result`.

In general, 
```
return someLogic();
```
is preferred to
```
const result = someLogic();
return result;
```

src/main/webapp/script.js
```diff
@@ -20,7 +20,12 @@ function createElementByTag(text, tag) {
 }
 
 function capitalizeFirst(string) {
```
zajonc-gugle: I think we can rename the function now, as it capitalizes all items.

src/main/webapp/script.js
zajonc-gugle: Let's remove this line.

src/main/webapp/parse.js
```diff
@@ -44,10 +46,13 @@ function sendData() {
             try {
                 const ingrHTML = document.getElementsByTagName("ul").item(0).children;
                 const list = document.getElementsByClassName("mw-parser-output")[0];
+                const ingredients = Array.from(ingrHTML).map(el => el.innerText);
                 const descriptionHTML = list.getElementsByTagName("p")[1];
+                var description = "";
+                if (descriptionHTML != null) {
+                    description = descriptionHTML.textContent ?? "";
+                }
```
zajonc-gugle: A more idiomatic way of writing this is
```
const description = descriptionHTML == null ? "" : descriptionHtml.textContent ?? "";
```
or, in a more modern way (not supported in older browsers, but I dont think we care)
```
const description = descriptionHTML?.descriptionHtml.textContent ?? "";
```
The `?.` is call an optional chain operator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Either way, in general it's preferred not to create variables with temporary values just to reassign them in an if, e.g.
```
var x = "default_value";
if (some_condition) {
    x = "non_default_value;
}
```
It's more common to assign the value to the result of an inline `if`, in any form (`?`, `||`, `&&`, `?.`), e.g.
```
const x = some_condition ? "non_default_value" : "default_value";
```

src/main/webapp/parse.js
zajonc-gugle: Yeah, let's use `const` here.

zajonc-gugle (changes requested):

PR #37, opened 2020-09-15
Force SSL

grenlayk:
Added `web.xml` file to forse https


polinasand (approved):

zajonc-gugle (approved):
I've never done this, so let's wait for Bartek's approval.

bprusak (approved):
As far as I know, that's the way it should be configured, but outside of reading about it I have never tried it either. Let's try now.

PR #46, opened 2020-09-18
Update search algo

polinasand:
Now list of results displays in descending order, first recipe has the biggest number of keywords from request, next has less or equal, and so on.

### DIFF
--- a/src/main/java/com/google/sps/data/Meal.java
+++ b/src/main/java/com/google/sps/data/Meal.java
@@ -15,6 +15,7 @@
 package com.google.sps.data;
 
 import java.util.ArrayList;
+import java.util.List;
 
 public class Meal {
     private final Long id;
@@ -49,5 +50,29 @@ public class Meal {
 
     public String getType() {
         return this.type;
+    }
+
+    /**
+     * Counts the number of keywords that are matching in recipe.
+     * @param params list of strings with keywords, that describe search request.
+     * @return amount of occurrences of keywords from params in this meal.
+     */
+    public int getFrequency(List<String> params) {
+        int frequency = 0;
+        for (String param : params) {
+            if (title.contains(param) ||
+                description.contains(param) ||
+                type.contains(param)) {
+                frequency++;
+                continue;
+            } 
+            for (String ingredient : ingredients) {
+                if (ingredient.contains(param)) {
+                    frequency++;
+                    continue;
+                }
+            }
+        }
+        return frequency;
     }
 }
--- a/src/main/java/com/google/sps/servlets/MealServlet.java
+++ b/src/main/java/com/google/sps/servlets/MealServlet.java
@@ -100,23 +100,22 @@ public class MealServlet extends HttpServlet {
         List<Meal> searchedMeal = new ArrayList<>();
         HashMap<Integer, List<Meal>> sortedMeal = new HashMap<>();
         for (Meal meal : meals) {
-            int count = isResultOfSearch(meal, params);
-            if (count > 0) {
+            int frequence = getFrequence(meal, params);
+            if (frequence > 0) {
                 List<Meal> list = new ArrayList<>();
-                if (sortedMeal.get(count) != null) {
-                    list = sortedMeal.get(count);
+                if (sortedMeal.get(frequence) != null) {
+                    list = sortedMeal.get(frequence);
                 }
                 list.add(meal);
-                sortedMeal.put(count, list);
+                sortedMeal.put(frequence, list);
             }
         }
         for (int i = params.size(); i > 0; i --) {
-            if (sortedMeal.get(i) == null) {
-                continue;
-            }
             List<Meal> list = sortedMeal.get(i);
-            for (Meal meal : list) {
-                searchedMeal.add(meal);
+            if (list != null) {
+                for (Meal meal : list) {
+                    searchedMeal.add(meal);
+                }
             }
         }
         response.setContentType("application/json");
@@ -204,23 +203,23 @@ public class MealServlet extends HttpServlet {
         return (Meal)mealList.get(index);
     }
 
-    private int isResultOfSearch(Meal meal, List<String> params) {
-        int counter = 0;
+    private int getFrequence(Meal meal, List<String> params) {
+        int frequence = 0;
         for (String param : params) {
             if (meal.getTitle().contains(param) ||
                 meal.getDescription().contains(param) ||
                 meal.getType().contains(param)) {
-                counter ++;
+                frequence++;
                 break;
             } 
             for (String ingredient : meal.getIngredients()) {
                 if (ingredient.contains(param)) {
-                    counter ++;
+                    frequence++;
                     break;
                 }
             }
         }
-        return counter;
+        return frequence;
     }
     private String getParameter(HttpServletRequest request, String name, String defaultValue) {
         String value = request.getParameter(name);
--- a/src/main/java/com/google/sps/servlets/MealServlet.java
+++ b/src/main/java/com/google/sps/servlets/MealServlet.java
@@ -32,6 +32,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Random;
 import java.util.stream.Collectors;
@@ -97,8 +98,24 @@ public class MealServlet extends HttpServlet {
         // Every object Meal adds if it contains at least one keyword in at least one field 
         // (type, title, description, ingredients).
         List<Meal> searchedMeal = new ArrayList<>();
+        HashMap<Integer, List<Meal>> sortedMeal = new HashMap<>();
         for (Meal meal : meals) {
-            if (isResultOfSearch(meal, params)) {
+            int count = isResultOfSearch(meal, params);
+            if (count > 0) {
+                List<Meal> list = new ArrayList<>();
+                if (sortedMeal.get(count) != null) {
+                    list = sortedMeal.get(count);
+                }
+                list.add(meal);
+                sortedMeal.put(count, list);
+            }
+        }
+        for (int i = params.size(); i > 0; i --) {
+            if (sortedMeal.get(i) == null) {
+                continue;
+            }
+            List<Meal> list = sortedMeal.get(i);
+            for (Meal meal : list) {
                 searchedMeal.add(meal);
             }
         }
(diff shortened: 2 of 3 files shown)


src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -97,8 +98,24 @@ private void getMealList(HttpServletRequest request, HttpServletResponse respons
         // Every object Meal adds if it contains at least one keyword in at least one field 
         // (type, title, description, ingredients).
         List<Meal> searchedMeal = new ArrayList<>();
+        HashMap<Integer, List<Meal>> sortedMeal = new HashMap<>();
         for (Meal meal : meals) {
-            if (isResultOfSearch(meal, params)) {
+            int count = isResultOfSearch(meal, params);
```
grenlayk: `count` of what?

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -187,24 +204,23 @@ private Meal pickRandomMeal(List<Meal> mealList) {
         return (Meal)mealList.get(index);
     }
 
-    private Boolean isResultOfSearch(Meal meal, List<String> params) {
+    private int isResultOfSearch(Meal meal, List<String> params) {
```
grenlayk: It's kind of strange that `isResultOfSearch` returns `int` value. Probably we need to rename it.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -97,8 +98,24 @@ private void getMealList(HttpServletRequest request, HttpServletResponse respons
...
+                }
+                list.add(meal);
+                sortedMeal.put(count, list);
+            }
+        }
+        for (int i = params.size(); i > 0; i --) {
+            if (sortedMeal.get(i) == null) {
+                continue;
+            }
+            List<Meal> list = sortedMeal.get(i);
```
grenlayk: What about 
```
List<Meal> list = sortedMeal.get(i);
if (list != null) {
  for (Meal meal : list) {
    searchedMeal.add(meal);
  }
}

```

not to call `sortedMeal.get(i)` twice and make code a little bit more readable

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -187,24 +204,23 @@ private Meal pickRandomMeal(List<Meal> mealList) {
...
         for (String param : params) {
-            if (meal.getTitle().contains(param)) {
-                return true;
-            } 
-            if (meal.getDescription().contains(param)) {
-                return true;
+            if (meal.getTitle().contains(param) ||
+                meal.getDescription().contains(param) ||
+                meal.getType().contains(param)) {
+                counter ++;
```
grenlayk: I have never seen `++` operator with spacing between variable (it always was smth like `counter++`). Are you sure that it is right according to code style?

polinasand: I met both `i++` and `i ++`, but let first will be here.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -97,9 +98,24 @@ private void getMealList(HttpServletRequest request, HttpServletResponse respons
...
-            if (isResultOfSearch(meal, params)) {
-                searchedMeal.add(meal);
+            int frequency = getFrequency(meal, params);
+            if (frequency > 0) {
+                List<Meal> list = new ArrayList<>();
+                if (sortedMeal.get(frequency) != null) {
+                    list = sortedMeal.get(frequency);
+                }
+                list.add(meal);
+                sortedMeal.put(frequency, list);
```
bprusak: We should change the order of things that are going on here.
That's because in Java `List list = mymap.get(key);` will get you not a copy of the list, but reference to that list - if you modify the list in any way (with `.add()` etc) then the original object gets updated - the one that also happens to still be in the map.
Therefore we can do the following:
1. Make sure that there's a list under that key with `.putIfAbsent(new ArrayList<>());`
2. `.get()` the list (or use the result returned by `.putIfAbsent`, although since it sometimes returns `null` it'd be complicated)
3. Add an element to that list.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -97,9 +98,24 @@ private void getMealList(HttpServletRequest request, HttpServletResponse respons
...
+            }
+        }
+        for (int i = params.size(); i > 0; i --) {
```
bprusak: If you had a `TreeMap` instead of a `HashMap` here, you would be able to just go through the map using `.descendingMap()` or `.descendingKeySet()`.
`TreeMap` keeps all elements in order (i.e. if you put entries under keys `1`, `7`, `5` and `9`, then they would be sorted in ascending order in the map, unlike the pseudo-random order of a `HashMap`) at a cost of making all insert and get operations in `O(log(n))` time instead of `O(1)`. These structures are great for sorting operations like this search result sorting.
If you have the time, please consider trying `TreeMap` here instead of `HashMap` (and then replacing the `for (int i = ....)` with a loop over `mymap.descendingMap()`, because otherwise the whole endeavour won't have any benefits; also looping over the map guarantees the value can't be `null`, so one less check to make!).

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -97,9 +98,24 @@ private void getMealList(HttpServletRequest request, HttpServletResponse respons
...
+            List<Meal> list = sortedMeal.get(i);
+            if (list != null) {
+                for (Meal meal : list) {
+                    searchedMeal.add(meal);
```
bprusak: Use `searchedMeal.addAll(list)` to add all elements of `list` to `searchedMeal` without the `for` loop.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -97,9 +98,24 @@ private void getMealList(HttpServletRequest request, HttpServletResponse respons
         // Every object Meal adds if it contains at least one keyword in at least one field 
         // (type, title, description, ingredients).
         List<Meal> searchedMeal = new ArrayList<>();
+        HashMap<Integer, List<Meal>> sortedMeal = new HashMap<>();
         for (Meal meal : meals) {
-            if (isResultOfSearch(meal, params)) {
-                searchedMeal.add(meal);
+            int frequency = getFrequency(meal, params);
```
bprusak: Add a test that checks whether an entry with more matches is returned at the beginning of the results list, favoured over other entries.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -187,24 +203,23 @@ private Meal pickRandomMeal(List<Meal> mealList) {
...
-            if (meal.getTitle().contains(param)) {
-                return true;
-            } 
-            if (meal.getDescription().contains(param)) {
-                return true;
+            if (meal.getTitle().contains(param) ||
+                meal.getDescription().contains(param) ||
+                meal.getType().contains(param)) {
+                frequency++;
+                break;
```
zajonc-gugle: This may be just my opinion, but I tend to avoid `break` and replace them with more explicit data flow operations whenever possible (like `if ... else`, for example).
In Java, in most cases you can refactor your code not to use loops (and, in consequence, `break`) by using the [Stream](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html) feature. I understand why it may feel like an overcomplicated approach (a `for` loop seems simpler, at a first glance), but in some cases it really pays off. (long story short, it is, in my opinion, far easier to write a for loop with a bug than a `Stream` chain with a bug)

Feel free to ignore the Streams suggestion, though. I just wanted to point out that there's usually many approaches to solve the same problem, and it's often useful to consider all of them.

However, I believe this block of code may have a tiny bug that may have been avoided had we used streams.

src/main/java/com/google/sps/servlets/MealServlet.java
```diff
@@ -187,24 +203,23 @@ private Meal pickRandomMeal(List<Meal> mealList) {
         return (Meal)mealList.get(index);
     }
 
-    private Boolean isResultOfSearch(Meal meal, List<String> params) {
+    private int getFrequency(Meal meal, List<String> params) {
```
zajonc-gugle: Now that the method has become a bit more complicated, I'd suggest adding a comment on top.

It wasn't necessary before, because the function name and return type were self explanatory, but that's no longer the case.
You can shortly describe what "frequency" is, in this context. Oftentimes a simple example serves as the best comment, e.g. 
```
getFrequency(
  new Meal(1, "Chocolate Cake", "A cake made of chocolate", ...),
  new List("chocolate", "cake")
) == <whatever you expect the result to be>
```
Pardon my Java, Bartek may know what the usual syntax for those "doctests" in Java is.

zajonc-gugle: Yet another suggestion to the `getFrequency` method.
Given that it's become fairly complicated, it could use some unit tests. We can test it through the MealServlet (and we should), but that's a bit cumbersome.
Another option would be to move the method to the `Meal` class. If we made it a public method of `Meal`, we could:
- distribute the code more evenly among files (it's usually a good practice to have many, relatively small files, rather than one massive file that implements everything)
- make the `Meal` class more useful (anyone, not just MealServet, would be able to calculate it's word frequency)
- and, most importantly, we could unit test the `getFrequency` method in unit tests of the Meal class. Writing unit tests for `Meal` will be *much* simpler than writing tests for `MealServlet`, because the class is way simper.

zajonc-gugle (changes requested):

src/main/java/com/google/sps/data/Meal.java
```diff
@@ -50,4 +51,28 @@ public String getDescription() {
     public String getType() {
         return this.type;
     }
+
+    /**
+     * Counts the number of keywords that are matching in recipe.
+     * @param params list of strings with keywords, that describe search request.
+     * @return amount of occurrences of keywords from params in this meal.
+     */
+    public int getFrequency(List<String> params) {
```
zajonc-gugle, line 60: I'd suggest adding a unit test for this method in a MealTest.java file.
A test that would check the exact number returned by this method for some instances.

src/main/java/com/google/sps/data/Meal.java
```diff
@@ -50,4 +51,28 @@ public String getDescription() {
...
+            if (title.contains(param) ||
+                description.contains(param) ||
+                type.contains(param)) {
+                frequency++;
+                continue;
+            } 
+            for (String ingredient : ingredients) {
+                if (ingredient.contains(param)) {
+                    frequency++;
+                    continue;
```
zajonc-gugle, line 72: This `continue` will skip the remaining part of the inner loop and go to the next iteration. However, there's nothing happening *after* the `continue`, therefore this keyword does nothing (FWIU). This `continue` doesn't really do anything.
As a result, if a meal has 3 ingredients:
- chocolate flakes
- chocolate milk
- chocolate whatever

then as long as the word "chocolate" is not in the title / description / type, returned frequency will be 3. However, if the title *does* include the word "chocolate", we'll return early with `frequency` == 1. It looks like a tiny bug.

I believe you may have meant to break the outer loop instead. I think the simplest solution is to extract the second for loop to a method, e.g. `bool doesKeywordAppearInIngredients(String keyword)`. Then you could drop all those `continue` keywords from this method and just have a single `if`.

zajonc-gugle (changes requested):

PR #49, opened 2020-12-28
Vue example

zajonc-gugle:
It took me quite a while, but I've finally managed to find some time and provide a sample of a frontend app written 100% in Vue.js.
Long story short, I've been trying to rewrite what you guys have implemented in vanilla HTML + CSS + js and pack it into a single page app (except for a few places where I was a bit too lazy to do things exactly as I should). Some things are mocked out, so don't expect the whole app to work - it's just a client making fake requests to a fake service.

Lessons learned:
- Vue is kinda cool
- CSS is just black magic
- I'm **really** bad at time management

It's probably not going to too useful, but in case you've come to hate frontend development because of all that manual HTML and CSS management, well, here's how it's done (in a simplified form) these days. Some things are much easier this way, there's still plenty of things to drive people crazy, but, all in all, it's a bit more manageable and, most importantly, you can write unit tests.

Either way - don't worry too much about this PR - it can hang here forever. I just had some spare time during Christmas so I've decided to finally send a PR I've been talking about.

P.S. Happy new Year!


grenlayk:
Nice to hear from u!
I will look through it during winter holidays (at least, I hope so)

P.S. Happy New Year to you too!)

